PR #59, opened 2017-05-27
feat(draft): add -o/--dest flag

technosophos:
Provide 'draft create' with a flag to allow a directory to be specified
at the CLI, e.g. 'draft create -a foo -o mydir/'

This is useful for tooling that needs to create multiple draft projects
without CDing into a directory first. *cough* VS Code plugin *cough*


bacongobbler:
LGTM

PR #86, opened 2017-06-01
docs(install): adding Azure to token generation docs

kris-nova:
For those using the Azure Container Registry ;)


squillace:
@kris-nova thanks!

PR #92, opened 2017-06-01
Adding a typescript pack

arschles:
Adding a default pack for Typescript. I tried to assume as little as I could of the typescript configuration. Assumptions I had to make:

- That the codebase can be built with `tsc`, and doesn't need some higher level tool (I didn't add support for `ts-node`)
- That the user has all their dependencies installable with `npm` (i.e. in their `package.json`)
- That the user has a complete `tsconfig.json` that doesn't need to be generated


bacongobbler:
@arschles just an FYI that if you want to pick this up again, the code for the default packs are now stored [here](https://github.com/Azure/draft/tree/master/packs) and are generated thanks to go-bindata.

bacongobbler:
ping @arschles, any time you might have to take a look at this again? If not no worries, I can pick this up at some point. I'm looking to add more language packs to draft in the next release.

arschles:
@bacongobbler yep, I'm back at it right now - sorry for the delay 😄

packs/typescript/NOTICE
```diff
@@ -0,0 +1,9 @@
+MIT License:
+
+Copyright (C) 2011-2015 Heroku, Inc.
```
bacongobbler: did this code actually come from Heroku or did you create the `detect` script on your own? I don't see a typescript buildpack... https://github.com/heroku/heroku-buildpack-typescript is a 404 for me.

arschles: Nope, it didn't come from heroku. Good catch - I'll delete

alexpilotti:
Except for `RUN npm install -g typescript` in the Dockerfile, this pack should be similar / identical to the node one. Wouldn't it make sense to have a single pack that covers both and include the typescript example?
Adding something like the following to the node pack Dockerfile could do IMHO: 

`RUN if [ -f tsconfig.json ]; then npm install -g typescript; fi`

alexpilotti:
Adding to the comment above, since Visual Studio 2017 does not generate a tsconfig.json in the typescript node.js projects, the Dockerfile needs quite a few changes to address that as well. After playing with it a bit, I got:

```
FROM node:6-onbuild
RUN npm install -g typescript
RUN npm install --save @types/node
RUN npm install
RUN find . -name "*.ts" -not -path "./Scripts/*" | xargs tsc
EXPOSE 8080
ENV port 8080
ENV PORT 8080
CMD ["npm", "start"]   
```
An alternative approach consists in running msbuild (requiring a container image which includes it).

Considering that this PR clearly states that a tsconfig.json needs to be present, should we create a separate pack or should we try to address both use cases in a single pack?

bacongobbler:
TypeScript is a different language that eventually compiles down into javascript but has a completely different toolchain. The same could be said for CoffeeScript, Dart, and Elm. I think it's fine to have a completely separate language pack for a superset of Javascript.

TypeScript itself documents that ["The presence of a tsconfig.json file in a directory indicates that the directory is the root of a TypeScript project."](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html). I think @arschles is fine to assume that if there exists a tsconfig.json in the root directory, then it is a TypeScript app.

> Adding to the comment above, since Visual Studio 2017 does not generate a tsconfig.json in the typescript node.js projects, the Dockerfile needs quite a few changes to address that as well.

I would argue that it's not our responsibility to comply 1:1 with Visual Studio's project structure for a TypeScript app. I would say that it is our responsibility to comply to what the actual TypeScript community defines the project structure.

Now saying that, I do realize that there are a fair few [TypeScript trending repos on GitHub](https://github.com/trending/typescript) that do not define a tsconfig.json. Those that do not define one usually have some other file that we can check for:

- tsfmt.json
- tslint.json

If we check for the presence of one of those three files, I think we should be able to deploy most TypeScript apps.

If there's a better way to check for TypeScript apps, please let me know.

alexpilotti:
About detecting the language, since the heuristics involve include a lot of potentially complex guesswork, what about using [Linguist](https://github.com/github/linguist), which is what Github uses? There's a [Go porting](https://github.com/generaltso/linguist) which, if on par with the Ruby one, IMO would fit pretty well here.

alexpilotti:
About the build tooling, since tsconfig.json is not mandatory, there's also quite a lot of guesswork involved if we want to automate it (gulp, visual studio, custom build.sh, etc). 

For the time being, if typescript is detected but a tsconfig.json is not present, an option could be to just insert a note in the Dockerfile for the user to add their specific build commands, another one could be to just build every ts file in it: `RUN find . -name "*.ts" -not -path "./Scripts/*" | xargs tsc`

bacongobbler:
> There's a Go porting which, if on par with the Ruby one, IMO would fit pretty well here.

That's a good thought, but it's a CLI port rather than a client library. That would require some overhaul on the project side. However, that would be a good enhancement to the project. I'll file a new ticket for that.

bacongobbler:
#205 for tracking that discussion.

bacongobbler:
I'll leave it to @arschles to discuss implementation details with you over https://github.com/Azure/draft/pull/92#issuecomment-311227653. I'm not sure what's the best path forward on this one.

arschles:
Since I unfortunately have not had time to look at this in months, I am going to close this. I have added https://github.com/Azure/draft/issues/448 to track further discussion.

PR #104, opened 2017-06-05
(WIP) Added option to skip docker image push

### LINKED ISSUE #102: Allow for full local development with minikube
When using minikube, we can re-use the docker daemon that is running on the VM without having to use an external registry, as stated in the documentation:
 
> When using a single VM of Kubernetes, it’s really handy to reuse the minikube’s built-in Docker daemon; as this means you don’t have to build a docker registry on your host machine and push the image into it - you can just build inside the same docker daemon as minikube which speeds up local experiments.

Would it be possible to replicate this experience with draft and document how to do it?

rodcloutier:
This PR intents to add an option to skip the docker image push

Note: Will add proper tests if the design holds.

Fixes #102

### DIFF
--- a/docs/minikube.md
+++ b/docs/minikube.md
@@ -1,0 +1,74 @@
+## Minikube
+
+Draft requires a wildcard domain and an ingress controller. Here how to set everything
+up when using [Minikube][Minikube]
+
+### Setup the certificate
+
+Create a self-signed certificate using the following command. Don't forget to change
+the `*.change.me` value to your own wildcard domain that you have used to start Draft
+as mentionned in the [Installation Guide][Installation Guide].
+
+```
+$ openssl req -x509 -nodes -days 365 -newkey rsa:2048 -keyout ./tls.key -out ./tls.crt -subj "/CN=*.change.me
+```
+
+This will create two files, `tls.key` and `tls.crt`, that you will need to copy to the chart directory of
+your project.
+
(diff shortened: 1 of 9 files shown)


bacongobbler:
Given that

A) this is a hack that only works on a single-node cluster with
B) the Pod's pull policy is set to "IfNotPresent"

I don't think this is something we can reasonably support. This seems *very* edge-casey to me and could be fraught with lots of issues down the road. I'm curious to hear other people's thoughts on this one though.

docs/minikube.md
```diff
@@ -0,0 +1,106 @@
+## Minikube
+
+Draft requires a wildcard domain and an ingress controller. Here how to set everything
+up when using [Minikube][Minikube]
+
+### Setup the certificate
```
bacongobbler: This should not be required unless you're setting it up via kube-lego. Did something change in the default packs while I was away?

rodcloutier: I took the documentation from the installation documentation which was modified by commit https://github.com/Azure/draft/commit/9ff3a31e11a20d28a33ba8d8c8c9f18d26ee6a4f to include the wildcard domain information.

bacongobbler:
Unfortunately I'm going to say that this feature is not something we want to support. I really appreciate the effort you made to bring in this PR, but I don't want to bring this into draft.

However, let's keep the discussion rolling in #102 and see if we can come up with a different solution that will solve your needs, or if this use case needs to be solved though a separate tool.

Issue #95, opened 2017-06-02
Converting a set of packfiles into a defaultpacks/*.go


nomisbeme:
I have a working pack (under ~/.draft/packs) and I'd like a convenient way to generate the corresponding defaultpacks/*.go boilerplate (or eliminate the need for them.)

bacongobbler:
Can you elaborate more on this? I'm not sure I quite understand.

ultimateboy:
First of all, it's awesome to hear that you've made your own pack! Out of curiosity, for what language is the pack you created?

As you've noticed, the defaultpacks are hardcoded at this point in time. We are hoping to implement a draft packs manager plugin at some time in the future which will probably change how we handle the defaultpacks: https://github.com/Azure/draft/issues/2

At this time, there's no convenient way to generate the .go boilerplate code for defaultpacks. If you need help converting your pack into a defaultpack for a PR, please open another issue and we'll help you out!

nomisbeme:
I'm working on cleaning up my pack for Swift:
http://asciinema.org/a/aqj9x08jy0dj

@technosophos was working on the hardcoding of default packs last night at hackfest last night. He asked me to create this issue so we could track the work.

technosophos:
I am working on a `go-bindata`-based implementation of what is there so that we can lower the barrier for maintaining the existing packs.

nomisbeme:
Awesome, happy to give that a whirl when it's ready.
I sorted out my container image issues this morning so that's my last dependency on a PR for https://github.com/Azure/draft/issues/96

PR #108, opened 2017-06-05
feat(packs): use go-bindata to pack builtin packs

technosophos:
This replaces the hard-coded built-in packs with a generator that uses
go-bindata to read the `packs/` directory and generate an internal
representation that is compiled into the binary.

It adds the make target `generate`, which is called by the build and
cross-compile steps.

Doing things this way has the following advantages:

- Managing built-in packs is much easier (see #95)
- The packed binaries are compressed, so overall memory footprint is
  smaller

The main disadvantage is that it is harder to share files across the
packs. However, it is likely that these paks would diverge over time
anyway.


technosophos:
@msftclas okay

technosophos:
@bacongobbler I fixed the commit, but the bot does not seem to re-run.

bacongobbler:
Lemme try something...

technosophos:
Okay, given that we all know that at this point the error is with the bot, can we just manually change the label?

bacongobbler:
done. Merge when ready.

bacongobbler:
Would you mind rebasing this when you have the time? I merged #110 which means that this change will need to go into the bindata as well. Sorry!

technosophos:
Rebased with all of the changes to the build packs

PR #110, opened 2017-06-06
fix(defaultpacks): set basedomain

errordeveloper:
basedomain is refered to in all of the ingress templates, but unset
in default values file, which results in regex validation error for
the ingress object


bacongobbler:
the basedomain is present in both forms of values.yaml, so it should be available in all charts:

https://github.com/Azure/draft/blob/3fad7ef57ce9ed198649bef38023be3860792d24/chart/values.yaml#L4
https://github.com/Azure/draft/blob/3fad7ef57ce9ed198649bef38023be3860792d24/cmd/draft/installer/install.go#L25

However I think this is an overall okay change.

PR #118, opened 2017-06-06
inform user when pack already exists

### LINKED ISSUE #6: draft implies packs being overwritten on `draft init`
While testing a change to the default python pack I noticed that `draft init -c` did not re-create the pack if it already exists, although it implies it did:

```
$ draft init -c
Creating pack python
...
Happy Sailing!
```

The CLI should inform the user that the pack will not be overwritten.

bacongobbler:
closes #6

### DIFF
--- a/cmd/draft/init.go
+++ b/cmd/draft/init.go
@@ -204,9 +204,13 @@ func ensurePacks(home draftpath.Home, out io.Writer) error {
 		return err
 	}
 	for packName, files := range all {
-		fmt.Fprintf(out, "Creating pack %s\n", packName)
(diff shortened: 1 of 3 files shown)


michelleN:
This doesn't seem like the best experience. We should probably just mentioned we're skipping and not return an err. What do you think? @bacongobbler 
```
draft [bacongobbler-6-draft-init*]$ bin/draft init
Creating pack python...
Pack python already exists. Skipping!
Error: pack already exists
```

bacongobbler:
Ah yeah that was the intention. Good catch.

bacongobbler:
fixed

PR #144, opened 2017-06-12
fix(chart): use linux node selector on draftd deployment

### LINKED ISSUE #143: Use annotations to avoid deployment on Windows nodes for hybrid clusters
If you have an hybrid cluster (windows + linux pools), draftd will end up on a random node, and if it lands on a windows node, it will fail to pull the image. A simple annotation in the spec of the draftd will make sure it always lands on a linux node:

```
spec:
      nodeSelector:
        beta.kubernetes.io/os: linux
```

but I'm not sure where to change it..

ultimateboy:
Fixes #143

### DIFF
--- a/chart/templates/deployment.yaml
+++ b/chart/templates/deployment.yaml
@@ -46,3 +46,5 @@ spec:
       - name: docker-socket
         hostPath:
           path: /var/run/docker.sock
+      nodeSelector:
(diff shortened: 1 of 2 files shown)


ultimateboy:
Just a quick cross-reference to the same issue in kubernetes/helm: https://github.com/kubernetes/helm/pull/2544

PR #221, opened 2017-07-07
Fix "file perm denied error" issue

lastcoolnameleft:
sudo echo'ing can lead to "file permission denied error"


bacongobbler:
this section of the documentation no longer exists. Closing!

PR #241, opened 2017-08-07
feat(*): add plugin update cmd

michelleN:


cmd/draft/create_test.go
```diff
@@ -12,6 +12,7 @@ import (
 	"testing"
 
 	"github.com/Azure/draft/pkg/draft/draftpath"
+	helpers "github.com/Azure/draft/pkg/testing"
```
bacongobbler: just a thought... Would it be better to move the CopyTree function to `pkg/testing/helpers` if we're always aliasing this package to `helpers`?

michelleN: yeah I have no problem with that and it makes sense! I'll change it.

PR #239, opened 2017-08-02
remove detect script from packs, replace with github/linguist

bacongobbler:
closes #205 
closes #218

PR #236, opened 2017-07-31
make charts generated by draft helm-compatible

bacongobbler:
With this re-write of the `draft create` codebase, Draft will:

- detect a pack
- write the chart into chart/packname/

This differs from the existing behaviour because the current implementation writes the chart directly to chart/. This allows users to run `helm lint chart/packname` and more importantly, `helm package chart/packname`.

closes #227 

Some open questions about this design:

- Is the pack name what we want the chart name to be centered around? I'd love for it to be the application name (usually the root directory of the application directory), but that was eventually shot down when we needed to auto-generate names for Draft applications.

PR #244, opened 2017-08-09
remove package api since grpc support.

fibonacci1729:


bacongobbler:
I'm seeing the same error in #236 and #239 so maybe there's a test bug in master that we missed before merging in #237? I'll see what I can find and submit a PR.

bacongobbler:
... oh. It's because CI is using an older version of Go.

bacongobbler:
merge when ready.

PR #282, opened 2017-08-23
feat(cli-ux): add basic draft up ux

fibonacci1729:
draft_up


pkg/cmdline/cmdline.go
```diff
@@ -0,0 +1,156 @@
...
+		default:
+			fmt.Fprintf(cli.opts.stdout, "\r%s %c", m, s[i%len(s)])
+			time.Sleep(50 * time.Millisecond)
+			i++
+		}
+	}
+}
+
+func passStr(msg string) string {
+	const pass = "⚓"
```
bacongobbler: +💯

bacongobbler:
lmk when this is ready for testing!

bacongobbler:
looks like you also need to call `glide up` to make CI pass.

pkg/cmdline/cmdline.go
```diff
@@ -112,22 +112,20 @@ func progress(cli *cmdline, app, desc string, codes <-chan rpc.UpSummary_StatusC
...
-				case rpc.UpSummary_SUCCESS:
-					done <- fmt.Sprintf("%s: %s  (%.4fs)\n", cyan(app), passStr(desc), time.Since(start).Seconds())
-					return
-				case rpc.UpSummary_FAILURE:
-					done <- fmt.Sprintf("%s: %s  (%.4fs)\n", cyan(app), failStr(desc), time.Since(start).Seconds())
-					return
-				}
-			case <-cli.Done():
+		defer func() {
+			close(done)
```
bacongobbler: is there any technical difference between this and `defer close(done)`?

pkg/cmdline/cmdline.go
```diff
@@ -112,22 +112,20 @@ func progress(cli *cmdline, app, desc string, codes <-chan rpc.UpSummary_StatusC
...
+		}()
+		for code := range codes {
```
bacongobbler: Would it make more sense to just pluck a single status code from the channel?

fibonacci1729: perhaps, but it's structured this way to support increased verbosity, i.e. status code UpSummary_LOGGING, etc. in the future.

pkg/draft/context.go
```diff
@@ -65,3 +71,17 @@ func newAppContext(s *Server, req *rpc.UpRequest, out io.Writer) (*AppContext, e
...
+func getulid() string { return <-ulidc }
+
+// A channel which returns build ulids.
+var ulidc = make(chan string)
+
+func init() {
+	rnd := rand.New(rand.NewSource(time.Now().UTC().UnixNano()))
+	go func() {
+		for {
+			ulidc <- ulid.MustNew(ulid.Timestamp(time.Now().UTC()), rnd).String()
```
bacongobbler: Isn't this a little excessive on clock cycles to be constantly generating ULIDs even when nothing is requesting one?

What if we just changed the function to

```
func getulid() string {
    rnd := rand.New(rand.NewSource(time.Now().UTC().UnixNano()))
    return ulid.MustNew(ulid.Timestamp(time.Now().UTC()), rnd).String()
}
```

Or is there some technical reason this isn't possible?

bacongobbler:
I feel it would still be beneficial to display the helm release notes from the deploy after it succeeds. What do you think about that? Is that something we can tackle in this PR or is that something we'd need to follow up afterwards?

bacongobbler:
All that being said, this is 🔥

```
><> draft up
Draft Up Started: 'esteemed-mouse'
esteemed-mouse: Building Docker Image: ⚓  (57.0438s)
esteemed-mouse: Pushing Docker Image: ⚓  (8.0013s)
esteemed-mouse: Releasing Application: ⚓  (7.0567s)
esteemed-mouse: Build ID: 01BRAZ7AT5DK716R5HECARMQBT
```

pkg/draft/context.go
fibonacci1729: Makes no difference to me and as far as I can reason, technically, there isn't much overhead. Because ulidc is an unbuffered channel, ULIDs will only be generated whenever getulid is invoked, then subsequent ulid will then queue up in the channel for the next invocation. The only overhead is the nominal stack size for the goroutine. In fact, i'd be curious to measure the performance to determine if using `<-ulidc` is more performant than `getulid`.

fibonacci1729:
My plan is to handle that with the verbosity control option sent to dratfd in the next cycle. I still, however, believe that we should strive to keep the UI as minimal and uncluttered as possible, especially for things like release notes that may be optionally present. Ideally using the build id, info should display the relevant information pertaining to their app.

pkg/draft/context.go
bacongobbler: I'm constantly impressed with your code. This is actually much more performant than my suggestion from a very basic benchmark.

Current code as-is:

```
><> go test -bench=.
BenchmarkGetulid-8       3000000               443 ns/op
PASS
ok      github.com/Azure/draft/pkg/draft        1.983s
```

And my suggested changes actually make it about twice as slow.

```
><> go test -bench=.
BenchmarkGetulid-8        200000             10191 ns/op
PASS
ok      github.com/Azure/draft/pkg/draft        2.369s
```

I feel like I should be wearing a tinfoil hat for the rest of the day. Carry on!

bacongobbler:
I've also just given some thought about how removing the release notes ties into what @michelleN is working on with `helm connect` too, so it's probably better to hide that information anyways. 👍

PR #291, opened 2017-08-31
add PORT envvar to nodejs example

bacongobbler:
supercedes #196

I also renamed the example to "nodejs" to better reflect the example application.


michelleN:
looks good. just waiting till CI is green.

PR #317, opened 2017-09-12
ref(pkg/draft): remove ondraft val

michelleN:
We're not using this any where right now so removing to prevent confusion!


bacongobbler:
This was being used in another project, but it was to work around ingress rules. Now with `ingress.enabled` as a value, we can check that instead. LGTM

PR #344, opened 2017-09-20
port linguist's .gitattributes support to Go

bacongobbler:
This provides users with a way to override `draft create`'s language detection logic by manually specifying if the files should be included for detection or if it is a different language type than what is determined by linguist.

See https://github.com/github/linguist#overrides


bacongobbler:
I know of a few edge cases where this doesn't work properly, so I'll have to write a few more test cases around `linguist-documentation`, `linguist-vendored` and `linguist-generated`.

squillace:
+1. This is a linguist trick that we should support.

bacongobbler:
fixed and all ready to go.

PR #349, opened 2017-09-21
Clarify how to configure the base domain

radu-matei:
If you are not familiar with the base domain, there is nowhere stated that you must pass the `--ingress-enabled` flag to configure the ingress.


bacongobbler:
Since this is just docs it's good to merge. I'll fix the failing glide issues :)

Issue #168, opened 2017-06-20
break up `draft init` into a plugin architecture


bacongobbler:
Instead of writing a new provider or writing a new one for kubernetes cloud provider X, I would like to propose a plugin architecture.

The problem statement is:

- maintainers may not have an expansive knowledge base on every cloud provider, making PRs harder to approve/accept
- vendors may want to release their provider support on a schedule separate from Draft's release schedule
- vendors want to have core support on the best experience for their cloud
- vendors may want to tweak the bootstrap configuration to their liking

How I envision this working:

- user calls `draft init --provider=foo`
- an executable called `draft-init-foo` is executed, prompting the user for more information or fetching information from the local environment
- draft provides the bootstrapped plugin with the same option flags it was sent, e.g. `draft init --provider=azure --ingress-enabled=true` would pass those flags as `draft-init-azure --ingress-enabled=true`. Whether or not that plugin actually respects that option flag is up to the plugin
- the `draft-init-foo` plugin installs Draft, optimized specifically for the platform (e.g. on Azure it'll use ACR, on GKE it'll use GCR, minikube uses the registry add-on)

PR #364, opened 2017-09-27
[WIP] Add support for transparent GCR auth

errordeveloper:


errordeveloper:
Getting the token this way works, but I'm currently unable to set permissions correctly for push to work.

pkg/draft/registry_gcp.go
```diff
@@ -0,0 +1,57 @@
+package draft
+
+// based on https://github.com/weaveworks/flux/blob/1.0.1/registry/gcp.go
```
bacongobbler: Because this is inspired by an Apache 2 licensed project, we're going to need to comply as per the [Redistribution section of the license](https://github.com/weaveworks/flux/blob/54d128c1da9be395e85b7108a96c7af8a3a2db31/LICENSE#L89-L121).

errordeveloper: Yes, for the purpose of PoC, I've copied this code. I am happy to discuss this as we progress on the feature in general, and I will be able to find the best solution by liaising with my colleague who authored this code.

errordeveloper:
> Getting the token this way works, but I'm currently unable to set permissions correctly for the push to work.

Waiting to hear from @mikedanese and @roberthbailey on this.

errordeveloper:
> Getting the token this way works, but I'm currently unable to set permissions correctly for the push to work.

Ok, with @mikedanese help, I got the right pointers to this. I just need to pass `--scopes=default,storage-rw`. (I was previously misled by someone on the internet to modify the default GCE service account instead).

So that means the user will have to do the following:

1. create a cluster (making sure `storage-rw` scope is included)
2. `helm init`
3. `draft init` (setting registry URL to `gcr.io/<project-id>`, and leave username and password blank)

The code in this PR pretty much does this. We can discuss how the OAuth token is obtained, I've been told there is an official package I can probably use instead of copying code from Flux.

bacongobbler:
I'm starting to think we should probably pull this out into a full "Draft on Google" installation plugin a la #168 instead of trying to shoe-horn every cloud provider into the core of Draft. Initializing Draft on GKE is likely going to be very different than initializing Draft on Minikube or Azure, so we should make "install wizards" for those specific use cases.

There was also the issue with users deploying their Kubernetes clusters on Azure, using S3 and RDS for backing their cluster. As one could imagine, there was significant latency issues and it caused more gripe than it was worth.

Reason I say this is because I know the docker daemons running ACS are already authenticated to use other Azure services (like ACR), minikube is automatically configured to be able to communicate with the registry deployed locally via `minikube config set registry=1` and GCE VMs automatically have permissions to push to GCR, so making an optimized "install wizard" for each cloud provider would make sense here.

I know I said that we should take a more pragmatic approach to this, and I'm very sorry for wasting your time on this PR. I'm going to start spiking out #168 this afternoon and try to start whipping that into shape with Azure and Minikube and see where things go, then we can come back to this PR and see how we can integrate GCR/GKE support into `draft init`. There are a few ideas in here that would make sense to bring in though. I'll spike it out and let you know when I'm close.

errordeveloper:
@bacongobbler I understand your concerns. Please keep in mind that token needs to be refreshed periodically, so it cannot be passed in as a flag during installation. Also, pull from the registry is something that needs addressing more broadly. I don't know about Azure's registry, but ECR would definitely require a similar sort of thing.
Besides introducing GCR support, I believe that `getRegistryAuthAsPushOptions()` and `getRegistryAuthAsPullSecret()` introduced here make the code a little more readable (I've got pretty confused with the plain JSON version and base64-endcoded one, and more broadly `server.go` is kind of mixed bag of stuff at this point, so this minor refactoring seems to make sense).

bacongobbler:
I'm going to close this out for now until we come up with a better solution for authenticating against GCS, likely with a tailored solution for each cloud provider like what's proposed in #168 or something similar. We need to revisit the conversation around how we can support multiple docker container registry authentication strategies such as GCS's token refresh strategy, AAD with ACR, Quay robot accounts etc.

PR #371, opened 2017-09-28
pin cobra version to last known version

bacongobbler:
For some reason, when calling `make bootstrap` on Draft I see an error about certain packages cobra relies on were missing. Pinning to the version of cobra that's in glide.lock and calling `glide up` appears to fix the issue.


bacongobbler:
superseded via #385

PR #269, opened 2017-08-16
bump to helm 2.6 compatibility

bacongobbler:
also bumps up for kubernetes 1.7 support

PR #385, opened 2017-10-04
replace glide with dep

### LINKED ISSUE #380: switch to dep
As per [the latest release of glide](https://github.com/Masterminds/glide/releases/tag/v0.13.0), it would seem that it'd make sense to migrate over to [dep](https://github.com/golang/dep) as that is starting to be absorbed into the core Go toolchain [according to their roadmap](https://github.com/golang/dep/wiki/Roadmap).

bacongobbler:
closes #380


bacongobbler:
going to hold off on #269 until we figure out if we want to move forward with this.

mattfarina:
do it. come on... you know you wanna. ;-)

sdboyer:
any info i can provide to help? 😄

bacongobbler:
thanks @sdboyer, I think I got it but I appreciate the help! I just haven't had the time to rebase and kick this into gear until now :)

Gopkg.toml
```diff
@@ -0,0 +1,64 @@
...
+
+[[constraint]]
+  name = "golang.org/x/net"
+
+[[constraint]]
+  name = "github.com/ghodss/yaml"
+
+[[constraint]]
+  name = "k8s.io/helm"
+  version = "~v2.7.0"
```
michelleN: 🎉

Makefile
```diff
@@ -110,14 +108,14 @@ test-lint:
 
 .PHONY: test-unit
 test-unit:
-	$(GO) test $(GOFLAGS) -cover -run $(TESTS) $(PKG) $(TESTFLAGS)
```
michelleN: how come we're removing the `-cover` and `-run` flags?

pkg/linguist/linguist.go
```diff
@@ -11,7 +11,6 @@ import (
 
 	"github.com/Azure/draft/pkg/osutil"
 	log "github.com/Sirupsen/logrus"
-	"github.com/generaltso/linguist"
```
michelleN: how come we're removing the linguist package here?

Makefile
bacongobbler: hmm, I'm not 100% sure why I removed them. I think it was because I noticed a significant performance improvement while testing, but that's irrelevant to this PR. I'll add them back in and do some more digging.

pkg/linguist/linguist.go
bacongobbler: The main reason here is because of [the setup steps](https://github.com/generaltso/linguist#install) required to run generaltso/linguist. When you run `make`, it performs `go generate` to perform a `git` submodule checkout of github/linguist, runs a naive bayesian classifier against the samples which then generates the resultant Go code. I've taken that package and put it in pkg/linguist so developers won't have to do that any more.

Essentially it allows us to remove [this](https://github.com/Azure/draft/blob/5937caa7c2e678c25961b7d5699bea6fc2aca6e4/scripts/setup-apimachinery.sh#L27-L35) piece of hacky code. In its place, I've added a README to pkg/linguist so others know how to update pkg/linguist when we need to re-train the bayesian classifier against a newer version of github/linguist.

michelleN: yeah that makes sense! thank you

michelleN:
I'm thinking there might something funky going on with portforwarding on Draft's side. I can do a `helm install`, `helm list` just fine on its own.

pkg/draft/server.go
```diff
@@ -19,10 +19,10 @@ import (
 	"golang.org/x/net/context"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
+	"k8s.io/api/core/v1"
 	apiErrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	k8s "k8s.io/client-go/kubernetes"
-	"k8s.io/client-go/pkg/api/v1"
```
michelleN: how come we're switching to use kubernetes core here instead of client-go?

michelleN: how come we're switching to use kubernetes core here instead of client-go?

michelleN: Oh, I see this was done because of a dependency conflict with Helm?

bacongobbler: Yeah, it's due to upgrading to helm 2.7, which supports kube 1.8. We'll have to do the same thing when we bump to helm 2.8 (kube 1.9)

bacongobbler: You can see @adamreese's changes to helm for kube 1.8 support is similar: https://github.com/Azure/draft/blob/5937caa7c2e678c25961b7d5699bea6fc2aca6e4/scripts/setup-apimachinery.sh#L27-L35

Gopkg.toml
```diff
@@ -0,0 +1,72 @@
...
+
+[[override]]
+  name = "k8s.io/kubernetes"
+  version = "=v1.8.0"
+
+[[override]]
+  name = "github.com/imdario/mergo"
+  revision = "6633656539c1639d9d78127b7d47c622b5d7b6dc"
```
michelleN: 🎉

bacongobbler:
fixed the issue related to `draft init` being unable to connect to kubernetes. We needed to pin github.com/imdario/mergo to the same version as helm.

Makefile
bacongobbler: Looks like there have been performance improvements made to Go 1.9 since this PR was initially introduced. I've reverted this.

Issue #396, opened 2017-10-16
plugins need a cross-compatible install hook


bacongobbler:
Right now a plugin relies on a install hook to be installed in the proper location, such as a bash or python script. The issue with this workflow means that any user can write a script that satisfies their platform of choice, but it may not be compatible across all operating systems (Windows in particular). This install hook should be refactored in such a way that it is cross-compatible by default.

bacongobbler:
looking into Go's plugin package for a hot-swappable plugin system. Example here: https://github.com/campoy/golang-plugins

bacongobbler:
Poo.

> The plugin support is currently incomplete, only supports Linux, and has known bugs.

https://golang.org/pkg/plugin/

bacongobbler:
closed via #658

PR #402, opened 2017-10-17
implement SymlinkWithFallback

### LINKED ISSUE #401: [windows] symbolic links cannot be created as a regular user
My power is out so I cannot get the proper error right now, but last night I was testing `draft init` on a windows box and I constantly found my powershell prompt exiting with no logs showing the error. Running `draft init` from inside Visual Studio Code shows that the error indicated that `draft plugin install https://github.com/Azure/draft-pack-repo` did not succeed because [symbolic links can only be created when running as an administrator](https://docs.microsoft.com/en-us/windows/device-security/security-policy-settings/create-symbolic-links) (or when the user is given the correct permissions to create symbolic links). According to documentation it is [a security risk to grant regular users symlink permissions](https://www.slideshare.net/OWASPdelhi/abusing-symlinks-on-windows).

Perhaps we should use an alternative solution such as using `os.Rename` to move the downloaded plugin from the cache to the plugin directory. We'd no longer have the advantage of saving the plugin in the cache directory, but it'd be portable across the three targeted OSes (macOS, Windows, Linux)

In short, there's two actionable items here to look up:

1) See why `draft init` is causing powershell to close when failing to create symlink (uncaught error/upstream bug?)
2) Find alternative solutions to `os.Symlink` for Windows users

I made a short GIF to demonstrate the powershell 🐛. I'll update this report once my power's back up.

bacongobbler:
This function attempts to create a symbolic link, but falls back to os.Rename() when the user does not have permission to create symbolic links (e.g. Windows non-administrators).

closes #401

### DIFF
--- a/pkg/draft/pack/repo/installer/local_installer.go
+++ b/pkg/draft/pack/repo/installer/local_installer.go
@@ -1,11 +1,11 @@
 package installer
 
 import (
-	"os"
 	"path/filepath"
 
 	"github.com/Azure/draft/pkg/draft/draftpath"
 	"github.com/Azure/draft/pkg/draft/pack/repo"
+	"github.com/Azure/draft/pkg/osutil"
 )
 
 // LocalInstaller installs pack repos from the filesystem
@@ -42,7 +42,7 @@ func (i *LocalInstaller) Install() error {
 		return err
 	}
 
-	return os.Symlink(src, i.Path())
+	return osutil.SymlinkWithFallback(src, i.Path())
 }
 
 // Update updates a local repository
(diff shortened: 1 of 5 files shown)


bacongobbler:
Tested manually on Windows and moves the needle a little further. Now to fix #396...

```
PS C:\Users\mafishe\go\src\github.com\Azure\draft> .\bin\draft.exe init --client-only
Adding plugin https://github.com/Azure/draft-pack-repo...
Error: exec: "sh": executable file not found in %PATH%
Error: exec: "sh": executable file not found in %PATH%
```

Issue #327, opened 2017-09-14
.7 RC: no clickable link provided for draft connect on Linux


squillace:
Hi all, on Linux, with the rc 2 of .7, I get this. Because the "localhost:36165" doesn't have http:// in front of it, I can't click it to open the site. :-( I would prefer either we add that protocol prefix so that natural terminal clicks are offered, or else even just spawn the site. If I chose `Draft connect` it's fairly clear that the default browser should open it, right?

![image](https://user-images.githubusercontent.com/1422718/30408935-d2f18a14-98b6-11e7-8b5a-917ba3ea9c40.png)

squillace:
Elsewise... what an amazing improvement. AMAZING!!!

fibonacci1729:
1. I agree with the adding the "http(s)" prefix.
2. Would an optional `--open` flag (default to false) be sufficient?

bacongobbler:
I completely agree this would be a fantastic feature for web apps. The problem is that Draft does not deploy strictly web-facing applications, as some apps (like draftd or gitlab) are not HTTP based. In fact they're gRPC or SSH endpoints.

In kubernetes the port type can be either "TCP" (the default, such as all the example apps) or "UDP". We could attach that to the local address displayed on `draft connect`, but that still wouldn't be very useful to the webapp developer. That would get us closer to the target requirement, though.

radu-matei:
Reiterating through @squillace's comments on #422 (and posting in the correct issue...):

> 1. We could detect port 80 and prefix based on that.
> 2. We could spawn the location, but then add a "if it's a web app, click http://localhost:xxxx; otherwise, open the endpoint using the appropriate tool."
> 3. We could add some metadata in the draft.toml file for this, and prompt the user to switch it off/on?
> 4. ....uh....
I'm inclined to go with 2 or 3... thoughts?

bacongobbler:
I wouldn't mind option 2!

bacongobbler:
I'm leaning more and more towards just making the default protocol for `draft connect` set to `http://`. Even if the underlying application protocol isn't http, the user knows that and will handle the connection string accordingly. It's a nice enhancement for users who DO write web applications with Draft (which one could argue is probably the large majority of our userbase).

PR #422, opened 2017-10-24
Add 'http://' prefix to output message

stuartleeks:
Add `http://` prefix when outputting the localhost address with `draft connect`. This makes the link clickable in bash to make it even easier to get to browsing the application

### DIFF
--- a/cmd/draft/connect.go
+++ b/cmd/draft/connect.go
@@ -60,7 +60,7 @@ func (cn *connectCmd) run() (err error) {
 		return err
 	}
 
-	detail := fmt.Sprintf("localhost:%#v", connection.Tunnel.Local)
+	detail := fmt.Sprintf("http://localhost:%#v", connection.Tunnel.Local)
 	fmt.Fprintln(cn.out, "SUCCESS...Connect to your app on "+detail)
 
 	fmt.Fprintln(cn.out, "Starting log streaming...")


bacongobbler:
First off, thank you for this!

Did you happen to read my comment in https://github.com/Azure/draft/issues/327#issuecomment-329514088? I'm a little concerned putting this here and making the assumption that the pod we are connecting to is indeed a web application. It is completely possible that the application deployed by draft is *not* a web app (or listens on some other protocol, like UDP).

If the community prefers we do this for simplicity's sake then I'm all for it, but I'm concerned about the implication (the UI now shows it off as http-only... how can we make this applicable to non-http applications?)

squillace:
I personally appreciate the idea, as we've noticed the problem before, but the issue is that the app needs to signal to draft in some way that it's http. PLUS the clickable link thing only works depending on the console/shell/OS, it's not general. some click, some don't.

That said, what are the options?
1. We could detect port 80 and prefix based on that.
2. We could spawn the location, but then add a "if it's a web app, click http://localhost:xxxx; otherwise, open the endpoint using the appropriate tool."
3. We could add some metadata in the draft.toml file for this, and prompt the user to switch it off/on?
4. ....uh....

Absent a clear answer, I'm good with the slight hindrance in UX until we have other things of larger importance landed..... and I'm great with other suggestions..... each of the above has issues.

stuartleeks:
Thanks for the comments. I scanned the issues list yesterday, but sadly missed that issue!

bacongobbler:
I'm going to close this PR for the time being. Let's carry on the discussion over at #327 until we can come up with a happy solution that works for everyone :)

Thanks!

Issue #352, opened 2017-09-21
smarter detection on "start command"


bacongobbler:
Packs like the python pack are assuming that the application's start command is `python app.py`. That does not translate well for python 3 application out there. We need to come up with a better way to determine how the application is started.

PR #441, opened 2017-11-28
Adding some clarifications about the requirements of the application.

TopherGopher:
Problem Statement:
I attempted to use draft on a personal project and ran into several roadblocks that led me down a hole trying to debug. Had I just had this paragraph, I would have saved hours of debug time (since I believed the error was in my app).

Adding a small chunk of text clarifying the requirements for a Python application.

Ideally, it would be great for the help text to dynamically print depending on the application since the error messages were a little mystic, but that's down the line.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -10,6 +10,8 @@ There are multiple example applications included within the [examples directory]
 $ cd examples/example-python
 ```
 
+When creating your own application, note that your entrypoint python file MUST be named app.py and be present in the root directory. Any additional packages that the project requires should be specified in requirements.txt (also at the root of the project). The container expects the application to serve on port 8080, however it is not required to implement an endpoint to use the tool.
+
 ## Draft Create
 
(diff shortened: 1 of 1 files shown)


bacongobbler:
this looks good. :) we can eventually remove sections of this documentation once we have smarter detection in place for the default python pack in https://github.com/Azure/draft/issues/352. Thank you for the PR!

bacongobbler:
Would it be possible to throw this in a different document though? I would prefer to keep the scope of the getting started guide to *just* the example python application. Perhaps another document called "Further Reading" or something, which would describe next steps after the basic getting started guide?

TopherGopher:
I mean, I'm still really new with this project, but I kept looking for a document that was Python specific and Go specific. Really, at the end of the day, I wanted a troubleshooting doc ultimately.
But if you give me the fnames, I can move this wherever.

bacongobbler:
We don't have any language-specific documentation, but I think that would be the right way to go forward with this PR :)

PR #460, opened 2018-01-01
feat(cmd/draft): clarify plugin setup in init

michelleN:
+ make less verbose, add test


bacongobbler:
this looks like it was fixed via https://github.com/Azure/draft-pack-repo/pull/8, so it looks like we need to cut a new release of draft-pack-repo before coming back to this, right?

michelleN:
yep that's correct!

PR #463, opened 2018-01-03
Clojure support

kstrempel:


examples/example-clojure/test/hello_world/handler_test.clj
```diff
@@ -0,0 +1,14 @@
+(ns hello-world.handler-test
+  (:require [clojure.test :refer :all]
+            [ring.mock.request :as mock]
+            [hello-world.handler :refer :all]))
+
+(deftest test-app
+  (testing "main route"
+    (let [response (app (mock/request :get "/"))]
+      (is (= (:status response) 200))
+      (is (= (:body response) "Hello World"))))
```
bacongobbler: won't this test fail because `hello` returns "Hello Draft Kubernetes World"? We should either change the response to either "Hello World" to match this test, or change the unit test to check for "Hello Draft Kubernetes World" :)

kstrempel: Yes.. you are right. I was so excited that the draft was working. Forgot to fix the test. :-)

PR #475, opened 2018-01-05
ignore common java work directories

jstrachan:
to reduce the amount of data tarred up and POSTed to docker when building the docker image via the  command

### DIFF
--- a/packs/java/.dockerignore
+++ b/packs/java/.dockerignore
@@ -2,3 +2,6 @@ Dockerfile
 draft.toml
 chart/
 NOTICE
+target/
+work/
+.git/


jstrachan:
incidentally I wonder if the .dockerignore file would be used by draft for `draft up` - to avoid posting the files docker ignores? Or do we need to start adding language specific exclusions to `.draftignore` too?

bacongobbler:
Yes, [draft reads from .dockerignore](https://github.com/Azure/draft/blob/5277f1313597aee060e3d7eda7beca7913a14abb/pkg/build/build.go#L151-L191) to package the tarball. :)

.draftignore was implemented to ignore events from certain files when the watch flag is enabled.

PR #491, opened 2018-01-11
inprocess + configmaps storage for draftd

fibonacci1729:
This PR implements basic state storage for draft. The added `pkg/storage` provides both an inprocess memory store (`pkg/storage/inprocess`) and configmaps backed kubernetes storage (`pkg/storage/kube`). The idea here is to provide a common link to the distributed state for a draft application (e.g. helm release, docker image, etc.). The `pkg/storage.Object` type is used to hold references to this distributed state and is what is stored by the aforementioned storage implementations.


cmd/draftd/start.go
```diff
@@ -70,6 +74,7 @@ func newStartCmd(out io.Writer) *cobra.Command {
 	f.StringVar(&sc.registryURL, "registry-url", "127.0.0.1:5000", "the URL of the registry (e.g. quay.io, docker.io, gcr.io)")
 	f.StringVar(&sc.basedomain, "basedomain", "", "the base domain in which a wildcard DNS entry points to an ingress controller")
 	f.StringVar(&sc.tillerURI, "tiller-uri", "tiller-deploy:44134", "the URI used to connect to tiller")
+	f.StringVar(&sc.storageEngine, "storage", "inprocess", "storage engine draftd should use")
```
michelleN: will we change the default from inprocess to configmap in a later release?

cmd/draftd/start.go
```diff
@@ -119,9 +124,17 @@ func (c *startCmd) run() (err error) {
 		cfg.UseTLS = true
 		cfg.TLSConfig = tlscfg
 	}
+	switch c.storageEngine {
+	case "configmaps":
+		const namespace = "default"
```
michelleN: Should the namespace be taken from the `draft.toml` file instead of defaulting to "default"?

cmd/draftd/start.go
fibonacci1729: I have no issue with that. Perhaps, post or with `logs` next cycle?

cmd/draftd/start.go
fibonacci1729: Seems reasonable. Likely it will be pulled from the draftd chart value overrides after draft init.

bacongobbler:
Let's just make sure we capture [this comment](https://github.com/Azure/draft/pull/491#discussion_r163343161) as a follow-up ticket as discussed :)

Issue #467, opened 2018-01-04
`draft connect`


michelleN:
Creates a local tunnel to the application deployed on kubernetes, returning a localhost URL for debugging.

- This is a meta issue to track `draft connect`'s progress from alpha to beta. We'll finalize design and UX around draft connect here.
- There have been several feature and configuration requests for `draft connect` since its initial release
- One open question is, should the `draft connect` experience automatically be part of `draft up`. Feedback here from end users would be greatly appreciated
 
Items:
- [x] #429, #472 Tackling this issue of specifying a local port to connect to gives us the opportunity to build a way to consistently configure default connect options in the draft.toml file
- [x] #329 
- [x] #471

i386:
@michelleN playing with Draft today and its very nice! I have some thoughts around the ergonomics of `draft connect` and `draft up` for the "edit->publish->see" cycle.

I found it a little akward that I needed to restart `draft connect` and fish out the apps tunnel URL every time I executed `draft up`. What if `draft up` established the `draft connect` tunnel initially and then subsequent `draft up` commands reused the same tunnel connection? That way you could have `draft up` hooked into something like livereload, publishing changes to kubernetes, as the developer edits their source code.

bacongobbler:
I think that might be possible with #429. We're toying around with similar ideas though, yeah.

michelleN:
Thanks for the feedback @i386. Your feedback on workflow is very helpful. I've updated the meta issue description (above) with features that will get us to the workflow you've described.

radu-matei:
Added in #600

michelleN:
All aspects of this issue have been addressed. Thanks!

PR #492, opened 2018-01-11
Add environment selection flag to connect command

birdayz:
This PR adds the possibility to choose the environment through the --environment / -e flag in connect.
Since this is possible with "up", i would expect this from connect as well.

There are two issues with the current implementation:
- Not having the default "development" environment in the .toml config results in a panic when calling "draft connect", because draft falls back to the default "development" and reads it from the map - and afterwards accesses the nil return value. See here https://github.com/birdayz/draft/commit/22fbcaa020f2bdae3351f83c82bd212b265274d1#diff-19de32b90b575e618c86ae2c9c8a80d6L49
Fixed by this PR. 
- It is only possible to choose the environment in connect by setting the environment variable DRAFT_ENV. Since up allows the -e param, i added this to connect as well.

Since i was already touching that code, i moved the environment stuff to its own file, because i feel that it  does not fit into up.go anymore because it is used not only by up but also by connect now (and possibly more commands in the future).
I also added a test to cover the "environment variable" vs default selection.

Remark: i guess there is more room for improvement for the environment handling - at least from my point of view. if there is interest, we can further discuss this. For me it is not really clear and there are open questions - for example different environments with the same app name are literally the same (correct me if i am wrong here), unless they use a different namespace. I am not sure if this is clear for the user.

Relates to #467

### DIFF
--- a/cmd/draft/connect.go
+++ b/cmd/draft/connect.go
@@ -20,27 +20,29 @@ type connectCmd struct {
 }
 
 func newConnectCmd(out io.Writer) *cobra.Command {
-	cc := &connectCmd{
-		out: out,
-	}
+	var (
+		cc          = &connectCmd{out: out}
+		environment string
+	)
 
 	cmd := &cobra.Command{
 		Use:   "connect",
 		Short: "connect to your application locally",
 		Long:  connectDesc,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return cc.run()
+			return cc.run(environment)
 		},
 	}
+
 	f := cmd.Flags()
 	f.Int64Var(&cc.logLines, "tail", 5, "lines of recent log lines to display")
+	f.StringVarP(&environment, "environment", "e", defaultDraftEnvironment(), "the environment (development, staging, qa, etc) that draft will run under")
 
 	return cmd
 }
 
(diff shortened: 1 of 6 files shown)


bacongobbler:
Definitely a gap we missed. Nice catch, @birdayz. :)

bacongobbler:
> for example different environments with the same app name are literally the same (correct me if i am wrong here) unless they use a different namespace.

They can also have different feature flags enabled, different custom helm values etc. [DEP 006](https://github.com/Azure/draft/blob/master/docs/reference/dep-006.md) tries to clear this up a bit, but if there's gaps in the documentation we need to cover, let's open a new ticket and suss out the details! :)

birdayz:
@bacongobbler Yeah right! My concern was: if the app name is the same, it will override the chart of the other environment. Just by reading the name, i expected that just by having something in a different environment it is "isolated" - no matter if the app name exists in a different environment already. Is that current behavior intentional?

bacongobbler:
To a certain extent, yes. Calling `draft up` on an application that already exists will cause it to upgrade the chart and the image. I think we want to retain that behaviour unless it's confusing or there's a technical reason why we should not do that.

PR #494, opened 2018-01-12
Revert "remove docker-in-docker container"

### LINKED ISSUE #443: buildApp: buildImg error: Error response from daemon: client is newer than server (client API version: 1.29, server API version: 1.24)
This error occurs on AKS. The docker daemons are an older release.

bacongobbler:
This reverts commit 0f699021b60031e41d08cdff69ff1699f6db4cd0.

closes #443

### DIFF
--- a/charts/draftd/templates/deployment.yaml
+++ b/charts/draftd/templates/deployment.yaml
@@ -30,6 +30,8 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.namespace
+        - name: DOCKER_HOST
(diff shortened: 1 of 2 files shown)


bacongobbler:
Yeah we'll have to re-open them all back up, but then we can optimize for minikube and close them back our with #168.

bacongobbler:
> Do you have any other testing strategies here @bacongobbler ?

As long as it works on minikube then it should work on anything else as before. We could also test on Azure, but since we know v0.8.0 works fine on Azure we can assume this'll work there as well. :)

PR #489, opened 2018-01-10
added spring-boot example

jstrachan:
this is a basic spring boot app created via http://start.spring.io/ with web + actuator (which adds health checks)

can be used to demonstrate #476

PR #499, opened 2018-01-17
select the most popular language found with a pack

### LINKED ISSUE #476: pack detection should default to the most popular language found that has a pack
I just tried the pack detection on a simple spring boot java app via `draft create` and got
```
Error: Could not find a pack Q_Q
```

With debug I saw:
```
DEBU[0000] language: Shell percent: 36.794912 color: #89e051
DEBU[0000] language: Batchfile percent: 26.097383 color: #C1F12E
DEBU[0000] language: Java percent: 18.710249 color: #b07219
DEBU[0000] language: Maven POM percent: 12.777604 color:
DEBU[0000] language: Gradle percent: 3.878636 color:
DEBU[0000] language: GDB percent: 0.192889 color:
DEBU[0000] language: INI percent: 0.573454 color:
DEBU[0000] language: Coq percent: 0.130331 color:
```
Its quite common to have some shell scripts and sometimes not huge amounts of Java code / POM files etc.  

In this particular case its due to spring boot apps often including shell scripts to wrap invoking maven + gradle - which can be larger than the hello world Java source code ;)

So I wonder if a nicer approach would be to 
* iterate in size order through languages and find pick the most popular language which has a pack
* for larger languages (Shell and Batchfile in this example) log a warning that the languages were ignored due to there being no pack?

If folks are happy with the above I'd be happy to submit a PR. Would love draft to just work OOTB with any Spring Boot app generated from: http://start.spring.io/ for example!

bacongobbler:
closes #476 
also refs #489

### DIFF
--- a/cmd/draft/create.go
+++ b/cmd/draft/create.go
@@ -29,6 +29,10 @@ const (
 `
 )
 
+// ErrNoLanguageDetected is raised when `draft create` does not detect source
+// code for linguist to classify, or if there are no packs available for the detected languages.
+var ErrNoLanguageDetected = errors.New("no languages were detected")
+
 type createCmd struct {
 	appName string
 	out     io.Writer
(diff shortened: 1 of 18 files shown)


cmd/draft/create.go
```diff
@@ -166,25 +168,28 @@ func doPackDetection(home draftpath.Home, out io.Writer) (string, error) {
...
+				if file.IsDir() {
+					if strings.Compare(strings.ToLower(detectedLang.Language), strings.ToLower(file.Name())) == 0 {
+						packPath := filepath.Join(packDir, file.Name())
+						log.Debugf("pack path: %s", packPath)
+						return packPath, nil
+					}
 				}
 			}
 		}
+		fmt.Fprintf(out, "--> Draft could not find a pack for %s. Trying the next likeliest language...\n", detectedLang.Language)
```
michelleN: What do you think about something like "Draft could find a pack for %s with certainty. Trying to find the next likely language match..."

cmd/draft/create.go
```diff
@@ -29,6 +29,8 @@ const (
 `
 )
 
+var NoLanguageDetected = errors.New("No languages were detected.")
```
michelleN: Might want to change this to `ErrNoLanguageDetected` for lint guidelines

cmd/draft/create.go
bacongobbler: That's much better.

michelleN:
Other than the lint issue, this looks great @bacongobbler. Thanks!

Issue #486, opened 2018-01-09
e2e acceptance suite


michelleN:
This is a meta issue to track the progress of creating a suite of acceptance tests for testing new releases of Draft.

PR #505, opened 2018-01-22
Fix e2e tests

### LINKED ISSUE #504: `make test-e2e` fails to run
```sh
$ make test-e2e
./tests/e2e.sh
./tests/e2e.sh: line 12: ./tests/tools: No such file or directory
# testing apps that are expected to pass
switching to example-dockerfile-http/
Error: failed loading build context with env "nowatch": failed to load chart: open /Users/seanknox/go/src/github.com/Azure/draft/tests/testdata/good/example-dockerfile-http/charts: no such file or directory
checking that example-dockerfile-http v1 was released
Expected REVISION == 1, got
make: *** [test-e2e] Error 1
```

Some of these looks pretty easy to remedy; I'll take a look.

seanknox:
- [x] I noticed [`make test-e2e` would error before running any tests](https://github.com/Azure/draft/issues/504). This fixes those issues.
- [x] Also refactored some extract pass/fail logic into `testhelpers.sh` to allow the test script to explicitly call `pass` or `fail` functions instead of hard-coding into tests themselves. Long-term, we might be better served using an established test harness than our own maintained in bash.
- [x] Temporarily commented out failing `"testing watch and changing files"` section. Since they weren't passing anyway and the intention wasn't clear to me, I thought it'd be helpful to discuss whether they're helpful to keep.

Fixes #504.


tests/e2e.sh
```diff
@@ -6,10 +6,12 @@
 # installs all the apps under testdata and checks that they pass or fail, depending on their
 # success condition.
 
-cd $(dirname $0)
+TESTDIR="$(pwd)/tests"
```
bacongobbler: One thing with `make` is that you can run from a different directory via `make -C draft test-e2e`, which means assuming that we are running from $GOPATH/src/github.com/Azure/draft is not always going to be the case. `cd $(dirname $0)` should be the right way to set this up; $0 is the full path to the script ($GOPATH/src/github.com/Azure/draft/tests/e2e.sh) so `cd $(dirname $0)` *should* put us in the right directory. Do you have a specific error message that can help illuminate the issue?

seanknox: Following the docs at https://github.com/Azure/draft/blob/master/docs/contributing/hacking.md#test-your-changes, I was running `make test-e2e` which would error trying to source the test helpers after it had cd'ed:

```
$ make test-e2e
./tests/e2e.sh
./tests/e2e.sh: line 12: ./tests/tools: No such file or directory
```

bacongobbler: Ah okay! So maybe we can then just change line 12 to `. ./tools` and that should fix the whole problem. How does that sound?

seanknox: Sounds good, running `make test-e2e` works with this:

```
cd $(dirname $0)
PATH="$(pwd)/../bin:$PATH"

. ./testhelpers.sh

```

bacongobbler:
From discussions with @seanknox, it seems like we are good to go ahead with this PR, but in a follow-up it'd probably be best if we remove the `make test-e2e` targets and keep the scripts for historical purposes to drive a better acceptance suite via #486. Right now it's probably not best for new contributors to implement integration tests in a bash script. It's brittle, unused, and not hooked up to CI. Writing an acceptance test suite in bash is probably not the best idea either.

What does everyone else think about this proposal? ^

seanknox:
Another option: remove the e2e tests entirely for now until it can be given proper thought in #486. As e2e isn't (AFAIK) part of any Draft engineer's workflow and isn't part of CI, I'd opt to remove them for now.

I pushed up another commit removing e2e tests and referring documentation so y'all can take a look.

bacongobbler:
interesting dep failure though.

PR #514, opened 2018-01-24
fix master (dependencies)

fibonacci1729:


Gopkg.toml
```diff
@@ -51,6 +51,17 @@
 [[constraint]]
   name = "github.com/ghodss/yaml"
 
+[[constraint]]
```
bacongobbler: would you mind removing these constraints now that they're no longer necessary?

PR #516, opened 2018-01-25
Update --insecure-registry CIDR range

bacongobbler:
https://github.com/kubernetes/minikube/pull/2077 changed the default service cluster IP range to 10.96.0.0/12 from 10.0.0.0/24.

Without this fix, `draft up` on minikube will fail with the following error:

> buildApp: pushImg error: Get https://10.107.109.2/v1/_ping: dial tcp 10.107.109.2:443: i/o timeout


bacongobbler:
I've added back 10.0.0.0/24 so older minikube clusters are still supported.

Issue #429, opened 2017-11-06
Option to specify a local port for "draft connect"


isutton:
Currently `draft connect` binds the remote port in a random local port.

In order to help developers, it would be helpful to specify a local port in `draft connect`. Some use case:

* Front-end developer running the server part of his application remotely and locally developing the front-end bits, probably with webpack development server acting as a reverse proxy to the local port bound to the remote development Pod.

seanknox:
the helm kube client doesn't allow overriding the local port in a portforward tunnel, so the options for implementing are:

1. allow specifying local tunnel port in `k8s.io/helm/pkg/kube`
2. replace helm kube client with k8s.io/client-go (which allows specifying local port)

After talking with @michelleN we think the using client-go here makes sense in the short term. When `connect` moves into beta we can revisit this approach and see if that's still a good idea.

bacongobbler:
https://github.com/kubernetes/helm/pull/3213 fixes that upstream FYI. You might be able to take that work into consideration :)

seanknox:
Oh awesome and timely!

michelleN:
#521 adds a -c flag to specify which container to connect to. There was a question about what to do if there are multiple containers involved and I recommend we create tunnels to all containers by default. Does anyone have thoughts around that?

If we agree that we want to create tunnels to all containers by default then we'll want to take a step back and re-think this experience a bit. One option is to use the container port specified in the manifest as the port we'd want to connect to locally as well by default instead of choosing something random (so port 8080 on a container would map to 8080 on your machine only choosing something random is 8080 is unavailable on your machine). A way to override default behavior on the command line could be to add the `-p [port]:[port]` syntax like docker does in order to bind a desired port. We can also add a way to override default behavior using the `[port]:[port]` syntax in the draft.toml file (which just means adding more details to #472 )

This scenario only takes into account us connecting to one application pod at a time. This scenario gets much more hairy when we do `draft up` and `draft connect` in a different directory on a different service (or application) in which case we might either get more random port mappings or have to specify overrides. In my head, this an advanced use case and we'd have to determine first how users want to interact with their services. Would you do a `draft up` on many services but a `draft connect` on only one or are folks wanting the `draft up && draft connect` experience on multiple services at a time? If the latter, are folks okay with random port mappings in that scenario or configuring throught draft.toml and the command line?

radu-matei:
Closed #521 in favour of #528, which:

- adds draft connect -c flag
[...]

PR #521, opened 2018-01-26
[WIP] Add initial --container/-c flag for connect

radu-matei:
cc @michelleN


radu-matei:
So choosing which container to connect to works right now - what should be the default behaviour with multiple containers? Connect to the first one, or create tunnel to all containers (container ports) in the pod? (In this case where should the logs come from?)

radu-matei:
I would also tackle #429 here. Thoughts?

radu-matei:
Any ideas here?
Thanks!

michelleN:
Hey @radu-matei - Sorry for the delay! We were at a team offsite last week. 

Thanks so much for your contribution here. I'll review and test this shortly. I believe the default behavior should be to create a tunnel to all containers. This will help folks debug sidecar containers better!

radu-matei:
Ok - then where should the logs come from?
Right now, the logs come from the container you connect to.

radu-matei:
Closing in favor of #528, which removes old and not useful commits.

PR #525, opened 2018-01-31
java pack: `mvn` invocation should be non interactive (`--batch-mode`). Fix  #524

### LINKED ISSUE #524: java pack: maven build should be "non interactive" (`--batch-mode`)
Maven builds (`mvn clean package`) should run in non-interactive mode (`--batch-mode`) for cleaner logs and to ensure that the build will not be blocked.

cyrille-leclerc:
`mvn` invocation should be non interactive (`--batch-mode`). Fix  #524


bacongobbler:
thanks!

PR #529, opened 2018-02-05
fix(*): specify multiple insecure registries args

michelleN:


bacongobbler:
let's cut a v0.10.1 with this change. 👍

Issue #338, opened 2017-09-15
revisit how to get local url/port after draft connect streams logs


michelleN:
Currently, you get the initial localhost url/port right after starting up draft connect. It's hard to find that after logs have been outputted. Gotta scroll up real far back. We could potentially interleave the logs and a the a status like `Your connection is still alive. Connect your app on localhost:6003.`

bacongobbler:
This sounds like a great way to fix the problem, IMO. :+1:

Issue #329, opened 2017-09-14
draft connect should wait for pods to come up before establishing connection


squillace:
for all apps, the UX for the current `draft up` to `draft connect` has a gap: the starting of the containers. It doesn't take that long on minikube, end even shorter in the cloud with bigger machines, but I'm wondering whether it's not a better experience to check the pods for completion as part of the Building, Pushing, Releasing.... stages. Truth is, "Releasing" is really, "deploying the chart" and as a result, it returns SUCCESS when your chart is deployed but the containers are unready. This means that you are lead to _expect_ that you can immediately `draft connect`, but that fails for short while until the pods are running. 

Below, after Releasing comes back as success, on my minikube there's about ten seconds to wait before you can use `draft connect`. 

![image](https://user-images.githubusercontent.com/1422718/30410970-170a4b4e-98c3-11e7-8a98-4e68c4853aef.png)

bacongobbler:
In draft.toml you can set `wait` to true to make draft wait until the pods have deployed and are running. This is disabled by default due to issues with helm when the app fails to come up, but it would be great for demos to flip this flag.

related: https://github.com/Azure/draft/issues/200

bacongobbler:
The [Draft User Guide](https://github.com/Azure/draft/blob/master/docs/user-guide.md) goes into a little more detail on what can be tuned in draft.toml.

bacongobbler:
re-titled the issue to better reflect the work required here to fix this.

cc @michelleN

radu-matei:
Don't have experience with the Helm `--wait` flag, but can't we simply wait for the pods to be running by using the Kubernetes client directly?

Something in the lines [here](https://github.com/radu-matei/kube-exec/blob/master/kube_util.go#L183)

radu-matei:
How about this? 😄 
Would using the Kubernetes API to wait for pods make sense?

PR #536, opened 2018-02-07
Add draft -p flag to specify local port, draft connect to beta

radu-matei:
A couple of considerations:

- according to the discussion in #429 and on Slack, as well as to the [open Helm PR](https://github.com/kubernetes/helm/pull/3213), work to add local port for port forwarding has started upstream on Helm

- the main idea in the PR on Helm is not to have breaking changes on Helm packages until Helm 3 - the way the PR currently looks, it will at least break the `New` method - and has been stale for a couple of months.

- this PR basically modifies the Helm tunnel package into Draft in order to specify the local port to port forward

The basic idea would be that once the PR (in whichever form) merges back to Helm, we can switch the implementation of the tunnel and port forward back.
But until then, we can have the functionality that might unblock a bunch of people from using this with tools such as live reload and other stuff.

Right now, this can only specify the local port for a single container - should we allow for multiple as well?

Thoughts?
cc @bacongobbler, @michelleN

### DIFF
--- a/packs/clojure/charts/templates/deployment.yaml
+++ b/packs/clojure/charts/templates/deployment.yaml
@@ -5,6 +5,7 @@ metadata:
   labels:
     draft: {{ default "draft-app" .Values.draft }}
     chart: "{{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}"
+    build: {{ .Values.buildID }}
 spec:
   replicas: {{ .Values.replicaCount }}
   template:
@@ -12,6 +13,7 @@ spec:
       labels:
         draft: {{ default "draft-app" .Values.draft }}
         app: {{ template "fullname" . }}
+        build: {{ .Values.buildID }}
     spec:
       containers:
       - name: {{ .Chart.Name }}
--- a/packs/csharp/charts/templates/deployment.yaml
+++ b/packs/csharp/charts/templates/deployment.yaml
@@ -5,6 +5,7 @@ metadata:
   labels:
     draft: {{ default "draft-app" .Values.draft }}
     chart: "{{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}"
+    build: {{ .Values.buildID }}
 spec:
   replicas: {{ .Values.replicaCount }}
   template:
@@ -12,6 +13,7 @@ spec:
       labels:
         draft: {{ default "draft-app" .Values.draft }}
         app: {{ template "fullname" . }}
+        build: {{ .Values.buildID }}
     spec:
       containers:
       - name: {{ .Chart.Name }}
--- a/packs/go/charts/templates/deployment.yaml
+++ b/packs/go/charts/templates/deployment.yaml
@@ -5,6 +5,7 @@ metadata:
   labels:
     draft: {{ default "draft-app" .Values.draft }}
     chart: "{{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}"
+    build: {{ .Values.buildID }}
 spec:
   replicas: {{ .Values.replicaCount }}
   template:
@@ -12,6 +13,7 @@ spec:
       labels:
         draft: {{ default "draft-app" .Values.draft }}
         app: {{ template "fullname" . }}
+        build: {{ .Values.buildID }}
     spec:
       containers:
       - name: {{ .Chart.Name }}
--- a/packs/gradle/charts/templates/deployment.yaml
+++ b/packs/gradle/charts/templates/deployment.yaml
@@ -5,6 +5,7 @@ metadata:
   labels:
     draft: {{ default "draft-app" .Values.draft }}
     chart: "{{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}"
+    build: {{ .Values.buildID }}
 spec:
   replicas: {{ .Values.replicaCount }}
   template:
@@ -12,6 +13,7 @@ spec:
       labels:
         draft: {{ default "draft-app" .Values.draft }}
         app: {{ template "fullname" . }}
+        build: {{ .Values.buildID }}
     spec:
       containers:
       - name: {{ .Chart.Name }}
--- a/packs/java/charts/templates/deployment.yaml
+++ b/packs/java/charts/templates/deployment.yaml
@@ -5,6 +5,7 @@ metadata:
   labels:
     draft: {{ default "draft-app" .Values.draft }}
     chart: "{{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}"
+    build: {{ .Values.buildID }}
 spec:
   replicas: {{ .Values.replicaCount }}
   template:
@@ -12,6 +13,7 @@ spec:
       labels:
         draft: {{ default "draft-app" .Values.draft }}
         app: {{ template "fullname" . }}
+        build: {{ .Values.buildID }}
     spec:
       containers:
       - name: {{ .Chart.Name }}
--- a/packs/javascript/charts/templates/deployment.yaml
+++ b/packs/javascript/charts/templates/deployment.yaml
@@ -5,6 +5,7 @@ metadata:
   labels:
     draft: {{ default "draft-app" .Values.draft }}
     chart: "{{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}"
+    build: {{ .Values.buildID }}
 spec:
   replicas: {{ .Values.replicaCount }}
   template:
@@ -12,6 +13,7 @@ spec:
       labels:
         draft: {{ default "draft-app" .Values.draft }}
         app: {{ template "fullname" . }}
+        build: {{ .Values.buildID }}
     spec:
       containers:
       - name: {{ .Chart.Name }}
--- a/packs/php/charts/templates/deployment.yaml
+++ b/packs/php/charts/templates/deployment.yaml
@@ -5,6 +5,7 @@ metadata:
   labels:
     draft: {{ default "draft-app" .Values.draft }}
     chart: "{{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}"
+    build: {{ .Values.buildID }}
 spec:
   replicas: {{ .Values.replicaCount }}
   template:
@@ -12,6 +13,7 @@ spec:
       labels:
         draft: {{ default "draft-app" .Values.draft }}
         app: {{ template "fullname" . }}
+        build: {{ .Values.buildID }}
     spec:
       containers:
       - name: {{ .Chart.Name }}
--- a/packs/python/charts/templates/deployment.yaml
+++ b/packs/python/charts/templates/deployment.yaml
@@ -5,6 +5,7 @@ metadata:
   labels:
     draft: {{ default "draft-app" .Values.draft }}
     chart: "{{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}"
+    build: {{ .Values.buildID }}
 spec:
   replicas: {{ .Values.replicaCount }}
   template:
@@ -12,6 +13,7 @@ spec:
       labels:
         draft: {{ default "draft-app" .Values.draft }}
         app: {{ template "fullname" . }}
+        build: {{ .Values.buildID }}
     spec:
       containers:
       - name: {{ .Chart.Name }}
--- a/packs/ruby/charts/templates/deployment.yaml
+++ b/packs/ruby/charts/templates/deployment.yaml
@@ -5,6 +5,7 @@ metadata:
   labels:
     draft: {{ default "draft-app" .Values.draft }}
     chart: "{{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}"
+    build: {{ .Values.buildID }}
 spec:
   replicas: {{ .Values.replicaCount }}
   template:
@@ -12,6 +13,7 @@ spec:
       labels:
         draft: {{ default "draft-app" .Values.draft }}
         app: {{ template "fullname" . }}
+        build: {{ .Values.buildID }}
     spec:
       containers:
       - name: {{ .Chart.Name }}
--- a/packs/swift/chart/templates/deployment.yaml
+++ b/packs/swift/chart/templates/deployment.yaml
@@ -5,6 +5,7 @@ metadata:
   labels:
     chart: "{{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}"
     draft: {{ default "draft-app" .Values.draft }}
+    build: {{ .Values.buildID }}
 spec:
   replicas: {{ .Values.replicaCount }}
   template:
@@ -12,6 +13,7 @@ spec:
       labels:
         app: {{ template "fullname" . }}
         draft: {{ default "draft-app" .Values.draft }}
+        build: {{ .Values.buildID }}
     spec:
       containers:
       - name: {{ .Chart.Name }}


radu-matei:
Hi, @michelleN!
My thoughts precisely (see [this comment](https://github.com/Azure/draft/pull/537#issuecomment-363882956) on #537).

michelleN:
@radu-matei Just merged #537. That looks great. I love the experience there and your idea for <local_port:container>. Do you think we can merge that to the --container flag or do you think we need a separate flag?

radu-matei:
@michelleN - it could work using the `--container/-c` flag (just that the `--port/-p` flag seems a bit more natural and consistent with Docker).

My only mention for `draft connect -c container:8000` is that I might expose multiple ports for `container` (so I wouldn't make the mapping as `container:local_port`, but `container_port:local_port`.

Other than that, a single flag works for me.

michelleN:
oh right! multi-port. yes that's going to be an issue.

michelleN:
Definitely feel like we should use another flag instead of merging experiences and see how that goes. Maybe something like `--override-port local:<port>:<container>:<port:`. What do you think?

radu-matei:
Not necessarily an issue - here's what I have in mind - right now, `draft connect` creates tunnels to all containers in the pod. To simplify things, as you suggested, there should be a single flag (`-c` or `-p`) that would bind a local port (`-p` in my opinion, since we're talking about ports, and and this way it would be consistent with the Docker `-p` flag).

So as per the example in #537, `draft connect` outputs:

```
Connect to multi-port:80 on localhost:53010
Connect to multi-port:81 on localhost:53012
Connect to go:8080 on localhost:53014

```

Now, since all containers in a pod share the network interface, there is only one of each port, and we should be able to do: `draft connect -p <local_port>:<pod_port>`, and tunnel the specified pod port to my specified local port.

Should it also start tunnels to all other exposed ports? I would say so, in order to be consistent to the simple `draft connect` command.

And you should be able to configure the connect command in `draft.toml`.

So summarising so far:

- `draft connect` should first check in `draft.toml` for configuration
- if none is present,  and no flag is provided, connect to all ports
- if a `-p` flag is provided, tunnel to the specified local ports, but still create tunnel to the rest of the exposed ports in the container, as with `draft connect`.

How does this sound?

bacongobbler:
I feel like that would be a better experience overall. The UX overall appears to be much simpler in this regard:

```
$ draft connect

Connected to go:8000 on 127.0.0.1:1234
Connected to go:8001 on 127.0.0.1:1235
Connected to foo:8000 on 127.0.0.1:1236
Connected to foo:8001 on 127.0.0.1:1237
```

So the corresponding UX with --override-port would be

```
$ draft connect --override-port 8000:go:8000 --override-port 8001:go:8001 --override-port 8002:foo:8000 --override-port 8003:foo:8001
```

And with `-c` in the mix:

```
$ draft connect -c go

Connected to go:8000 on 127.0.0.1:1234
Connected to go:8001 on 127.0.0.1:1235
```

The UX could look like this:

```
$ draft connect -c go --override-port 8000:go:8000 --override-port 8001:go:8001
```

We might even want to "strip" the `:go:` part in the last example there since it's already been specified with `-c`, though that may be more engineering trouble than its worth.

michelleN:
@bacongobbler What do you think about making `local` explicit?
Adding to your example: `$ draft connect -c go --override-port local:8000:go:8000

radu-matei:
At this point, the only thing I'm not happy about is `--override-port 8000:go:8000` - is it possible to have the same `containerPort` in the same pod for two different containers? (As the containers in a pod share the same network interface, what is the behaviour when setting the same port?)

Because in the end, `ForwardPort` creates tunnel to the pod, so back to whatever port the container exposes on the pod network interface.

Or am I missing something in pod networking and `ForwardPort`?

bacongobbler:
@michelleN that actually would be a nice feature. That way the names can become non-positional, which is somewhat of a nuisance to remember with `docker run -p`. e.g. `--override-port go:8000:local:8000` would be the same as `--override-port local:8000:go:8000`.



> is it possible to have the same containerPort in the same pod for two different containers?

Yes. Those are the port numbers internal to the container itself. The pod just can't have the same `hostPort` specified twice, as that is what is exposed via the pod's shared network namespace. `kubectl port-forward`'s APIs use the container's internal port so we don't need to worry about the pod's external ports at all.

A practical example:

```
><> cat deployment.yaml 
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  name: docker-registry
  labels:
    app: docker-registry
spec:
  replicas: 1
  selector:
    matchLabels:
      app: docker-registry
  template:
    metadata:
      labels:
        app: docker-registry
    spec:
      containers:
      - name: traefik
        image: nginx
        ports:
        - containerPort: 80
          hostPort: 80
      - name: registry
        image: registry
        ports:
        - containerPort: 5000
          hostPort: 80
      nodeSelector:
        beta.kubernetes.io/os: linux
><> k create -f deployment.yaml 
The Deployment "docker-registry" is invalid: spec.template.spec.containers[1].ports[0].hostPort: Duplicate value: "TCP//80"
```

radu-matei:
I am happy with the port configuration however you think it would be better (specific positions, as with Docker, or by specifying `local` and `remote`), and I would also include the configuration in `draft.toml` here:

- `draft connect` should first check in `draft.toml` for configuration
- if none is present,  and no flag is provided, connect to all ports
- if a `--override-port` flag is provided, tunnel to the specified local ports, but still create tunnel to the rest of the exposed ports in the container, as with `draft connect`.

Thoughts?

radu-matei:
Ok, here's what we have so far:

- since `ForwardPort` connects to the port exposed on the pod, it means we actually don't care for the container names when making the connection (we still output them when connecting, so it's easier to understand where you want to connect), so I removed the `--container/-c` flag - if you think it adds value, it can be brought back

- added an `--override-port/-p` flag that (in its current form) specifies the ports as `-p <local>:<remote>` (just like Docker), and an example would be `draft connect -p 8080:8080 -p 1234:80 -p 1235:81`

- also added an `override_ports` entry in the manifest, where you specify ports just as with the flag, space separated, so the command above is equivalent with the `draft.toml` config  `override_ports = "8080:8080 1234:81 1235:81"`

(- we can revisit the `local:<port>:remote:<port>` syntaxt, this was just easier to have a first draft - pun intended 😄  - working)

- if both the flag and config are used, the flag takes precedence

- also addressed #338  and outputs the connection message every 10 seconds

radu-matei:
Addressed #329 by using the Kubernetes API directly to wait for a pod to be in running state.

You can easily try this by using a deployment with init containers:
(also with multiple ports and containers, see above how to test it)

```yaml
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  name: {{ template "fullname" . }}
  labels:
    draft: {{ default "draft-app" .Values.draft }}
    chart: "{{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}"
spec:
  replicas: {{ .Values.replicaCount }}
  template:
    metadata:
      labels:
        draft: {{ default "draft-app" .Values.draft }}
        app: {{ template "fullname" . }}
    spec:
      initContainers:
      - name: init-container
        image: ubuntu
        command: ["/bin/sh"]
        args: ["-c", "sleep 20; echo Waited 20 seconds to be ready;"]
      containers:

      - name: multi-port
        image: radumatei/multi-port
        ports:
        - containerPort: 80
        - containerPort: 81

      - name: no-port
        image: ubuntu:14.04
        command: ["/bin/sh"]
        args: ["-c", "while true; do echo hello; sleep 5;done"]
        
      - name: {{ .Chart.Name }}
        image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
        imagePullPolicy: {{ .Values.image.pullPolicy }}
        ports:
        - containerPort: {{ .Values.service.internalPort }}
        resources:
{{ toYaml .Values.resources | indent 12 }}

```

radu-matei:
A status update here:

- updated the `toml` manifest to use dashes instead of underscores, so `wait_delay` became `wait-delay`, `override_ports` became `override-port` and removed the unused `wait` flag.

- added an `auto-connect` flag and config for `draft up` - there's one thing blocking it - if we try to auto-connect right after doing a `draft up`, and there is an existing Helm release of your application, Draft will try to connect before any pod is actually in `Terminating` state - which means it will find a ready pod and connect to it - then Helm will create a new release, killing the pod we are connected to, meaning that we have to reconnect.

- I tried to use the `wait` flag of Helm, but it didn't seem to work

- an idea would be to include the Helm release as a label on the pod spec, and when trying to connect to retrieve the latest release of the application and actually wait until a pod that satisfies all constraints is up (this part works, we only need to add the label).

Ideas here?
Also, feedback on the already implemented things in this PR would be great! 😄

squillace:
@radu-matei great strides here. I'll let the team comment on the PR in general as this week we were all in Portland for Helm summit, but I look forward to building/testing this weekend. GREAT usability improvements.

radu-matei:
@squillace - make sure to try with pods that have multiple ports, add the `override-port` flag and config and the waiting of pods before connecting.

Let me know what you think and if you have any questions 😄

cmd/draft/connect.go
```diff
@@ -43,7 +47,7 @@ func newConnectCmd(out io.Writer) *cobra.Command {
 	f := cmd.Flags()
 	f.Int64Var(&cc.logLines, "tail", 5, "lines of recent log lines to display")
 	f.StringVarP(&runningEnvironment, environmentFlagName, environmentFlagShorthand, defaultDraftEnvironment(), environmentFlagUsage)
-	f.StringVarP(&containerName, "container", "c", "", "name of the container to connect to")
```
michelleN: oh man I totally think we should bring the container flag back! I think it's a nice addition for people who don't care about ports but do care about only connecting to one container in their pod like a sidecar container. What do you think?

cmd/draft/connect.go
```diff
@@ -43,7 +47,7 @@ func newConnectCmd(out io.Writer) *cobra.Command {
...
+	f.StringSliceVarP(&overridePorts, "override-port", "p", []string{}, "specify a local port to connect to, in the form <local>:<remote>")
```
michelleN: 🎉

cmd/draft/connect.go
radu-matei: Sure, I'm open to it !
The only question is how to handle both `-c` and `-p` flags.

cmd/draft/connect.go
```diff
@@ -59,19 +63,34 @@ func (cn *connectCmd) run(runningEnvironment string) (err error) {
...
 	for _, cc := range connection.ContainerConnections {
 		for _, t := range cc.Tunnels {
 			err = t.ForwardPort()
 			if err != nil {
 				return err
 			}
-			fmt.Fprintf(cn.out, "Connect to %v:%v on localhost:%#v\n", cc.ContainerName, t.Remote, t.Local)
+			m := fmt.Sprintf("Connect to %v:%v on localhost:%#v\n", cc.ContainerName, t.Remote, t.Local)
+			connectionMessage += m
+			fmt.Fprintf(cn.out, m)
```
michelleN: 👍

cmd/draft/connect.go
michelleN: We'll have to parse the container flag first. We'll need to find which ports are open on that container and decide which local ports to forward based on which ones are open on that container. That logic should probably live in `pkg/draft/local`. What do you think?

michelleN:
hey @radu-matei - Would it be possible to move the `auto-connect` feature to a different PR?

radu-matei:
So far:

- brought back the `--container/-c` flag
- added more testing around connect in `pkg/draft/local`
- added more error handling for port mapping (return error when passing the same local/remote port multiple times)

From the way it looks, the `getsockopt: operation timed out ` error seems to be your cluster (Minikube?) related - I haven't had any issues with AKS so far - if anybody else can test this it would be great.

- I wouldn't move the `--auto-connect` flag just yet, as we're close to having a solid experience there - just added the build ID as label on deloyment and pods, which means we can safely wait until a pod with the right version of our app is running.

Thoughts?

cmd/draft/connect.go
```diff
@@ -92,8 +129,10 @@ func (cn *connectCmd) run(runningEnvironment string) (err error) {
 		os.Exit(0)
 	}()
 
-	<-stop
-	return nil
+	for {
+		fmt.Fprintf(cn.out, connectionMessage)
+		time.Sleep(10 * time.Second)
```
bacongobbler: I noticed that `draft connect` will continually spew the "Your connection is still active" message, so much that the majority of my logs are purely `draft connect` messages. Can we change the frequency to once every 5 minutes by default?

A future enhancement might be to display the message based on the number of logs being spewed out to the screen, but that kinda depends on the screen size.

cmd/draft/connect.go
```diff
@@ -59,19 +72,43 @@ func (cn *connectCmd) run(runningEnvironment string) (err error) {
...
+
 	for _, cc := range connection.ContainerConnections {
 		for _, t := range cc.Tunnels {
 			err = t.ForwardPort()
 			if err != nil {
 				return err
 			}
-			fmt.Fprintf(cn.out, "Connect to %v:%v on localhost:%#v\n", cc.ContainerName, t.Remote, t.Local)
+			m := fmt.Sprintf("Connect to %v:%v on localhost:%#v\n", cc.ContainerName, t.Remote, t.Local)
+			connectionMessage += m
```
bacongobbler: As of Go 1.10, there is a new [strings.Builder](https://golang.org/pkg/strings/#Builder) struct that makes this process easier. Not necessarily something that needs to be done in this PR but figured I should bring it up. :)

cmd/draft/connect.go
```diff
@@ -59,19 +72,43 @@ func (cn *connectCmd) run(runningEnvironment string) (err error) {
 		return err
 	}
 
-	connection, err := deployedApp.Connect(client, config, containerName)
+	var ports []string
+	if len(overridePorts) == 0 {
+		if deployedApp.OverridePorts != "" {
```
bacongobbler: Can we simplify the cyclical branch logic a little bit here?

```
if deployedApp.OverridePorts != "" {
	s := strings.Join(strings.Fields(deployedApp.OverridePorts), " ")
	ports = strings.Split(s, " ")
}

// --override-port takes precedence
if len(overridePorts) != 0 {
	ports = overridePorts
}
```

The end result should be the same, but the branch logic should be much simpler to grok.

Also see the below comment about making `OverridePorts` an array of strings instead of one long string. That would further simplify the code here as well as in draft.toml:

```
if len(deployedApp.OverridePorts) != 0 {
	ports = deployedApp.OverridePorts
}

// --override-port takes precedence
if len(overridePorts) != 0 {
	ports = overridePorts
}
```

and in draft.toml:

```
override-ports = ["8080:80", "8081:81"]
```

cmd/draft/testdata/create/generated/empty/draft.toml
```diff
@@ -2,6 +2,6 @@
   [environments.development]
     name = "myapp"
     namespace = "default"
-    wait = false
```
bacongobbler: how come the wait flag is being removed from draft.toml?

pkg/rpc/rpc.proto
```diff
@@ -15,6 +15,9 @@ service Draft {
 	// GetLogs returns the logs for a particular draft build.
 	rpc GetLogs(GetLogsRequest) returns (GetLogsResponse) {}
 
+	// GetLatestBuildID returns the latest build ID for a given application
+	rpc GetLatestBuildID(GetLatestBuildIDRequest) returns (GetLatestBuildIDResponse) {}
```
bacongobbler: Can we instead rename the RPC call as fetching the latest build instead of just its ID? e.g.

```
// GetLatestBuild returns the latest build for a given application
rpc GetLatestBuild(GetLatestBuildIDRequest) returns (GetLatestBuildResponse) {}
```

Even if functionally it's the same, this way we don't have to introduce new RPC functions any time we introduce a new field to a Build

In this sense, the response would then look something like this:

```
message GetLatestBuildResponse {
	string ID = 1;
}
```

Does that make sense?

pkg/kube/podutil/podutil.go
```diff
@@ -35,3 +42,30 @@ func GetPodCondition(status *v1.PodStatus, conditionType v1.PodConditionType) (i
 	}
 	return -1, nil
 }
+
+// GetPod waits for a pod with the specified labekl to be ready, then returns it
```
bacongobbler: typo: label*

pkg/draft/manifest/manifest.go
```diff
@@ -25,14 +25,15 @@ type Manifest struct {
...
-	WatchDelay   int      `toml:"watch_delay,omitempty"`
+	Name          string   `toml:"name,omitempty"`
+	BuildTarPath  string   `toml:"build-tar,omitempty"`
+	ChartTarPath  string   `toml:"chart-tar,omitempty"`
+	Namespace     string   `toml:"namespace,omitempty"`
+	Values        []string `toml:"set,omitempty"`
+	AutoConnect   bool     `toml:"auto-connect"`
+	Watch         bool     `toml:"watch"`
+	WatchDelay    int      `toml:"watch-delay,omitempty"`
+	OverridePorts string   `toml:"override-ports,omitempty"`
```
bacongobbler: This should be an array of strings instead of a string. That would make it easier to parse!

cmd/draft/connect.go
radu-matei: Sure, will update to 5 minutes.

radu-matei:
Closing as we're tracking each change in its own PR.
Thanks everyone for reviewing!

PR #537, opened 2018-02-07
Connect to every containerPort in pod

### LINKED ISSUE #535: If (any) container in pod does not expose ports, index out of range panic
If there is a container with no `containerPort` specified in the deployment file, because we try to connect by default to the first `containerPort` the container specifies, Draft will panic.

You can find the behaviour:

- [in the latest version (that connects to all containers in the pod)](https://github.com/Azure/draft/blob/master/pkg/draft/local/local.go#L79) - very likely to happen, especially for side containers

- [in the latest released version, that only connects to the first `containerPort` of the first container in the pod](https://github.com/Azure/draft/blob/486291360aeffa2fe0a131fbf0b4f3c601056b16/pkg/draft/local/local.go#L109) - it can still happen if I modify the deployment file and remove `containerPort`

In both cases, we should at least check that the container exposes at least one `containerPort` - I will look into it hopefully soon 😄 

Any thoughts?

radu-matei:
- fixes #535 
- starts tunnel to every `containerPort` in pod spec

Essentially, when doing `draft connect` with the deployment below, you get:

```
<tunnels to all ports of all containers>

Connect to multi-port:80 on localhost:53010
Connect to multi-port:81 on localhost:53012
Connect to go:8080 on localhost:53014

<logs from all containers, including ones without ports>
``` 

The `--container/-c` flag works as expected, creating tunnels only for all the ports of the selected container.

Best tried with the following deployment template, which has containers with 0, 1 or 2 defined `containerPort`

```yaml
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  name: {{ template "fullname" . }}
  labels:
    draft: {{ default "draft-app" .Values.draft }}
    chart: "{{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}"
spec:
  replicas: {{ .Values.replicaCount }}
  template:
    metadata:
      labels:
        draft: {{ default "draft-app" .Values.draft }}
        app: {{ template "fullname" . }}
    spec:
      containers:

      - name: multi-port
        image: radumatei/multi-port
        ports:
        - containerPort: 80
        - containerPort: 81

      - name: no-port
        image: ubuntu
        command: ["/bin/sh"]
        args: ["-c", "while true; do echo hello; sleep 1;done"]
        
      - name: {{ .Chart.Name }}
        image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
        imagePullPolicy: {{ .Values.image.pullPolicy }}
        ports:
        - containerPort: {{ .Values.service.internalPort }}
        resources:
{{ toYaml .Values.resources | indent 12 }}

```

### DIFF
--- a/pkg/draft/local/local.go
+++ b/pkg/draft/local/local.go
@@ -40,7 +40,7 @@ type Connection struct {
 
 // ContainerConnection encapsulates a connection to a container in a pod
 type ContainerConnection struct {
-	Tunnel        *kube.Tunnel
+	Tunnels       []*kube.Tunnel
 	ContainerName string
 }
 
@@ -63,9 +63,7 @@ func DeployedApplication(draftTomlPath, draftEnvironment string) (*App, error) {
 
 // Connect tunnels to a Kubernetes pod running the application and returns the connection information
 func (a *App) Connect(clientset kubernetes.Interface, clientConfig *restclient.Config, containerName string) (*Connection, error) {
-
 	var cc []*ContainerConnection
-
 	pod, err := getPod(a.Namespace, a.Name, clientset)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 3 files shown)


radu-matei:
For #429 and #536 to make sense, we should probably have `draft connect -p <local_port>:<containerPort> -p <another_local_port>: <another_containerPort>` - and maybe the logs should still come from all ports?

pkg/draft/local/local.go
```diff
@@ -141,21 +147,23 @@ func getPod(namespace, label string, clientset kubernetes.Interface) (*v1.Pod, e
...
-			port = int(c.Ports[0].ContainerPort)
+			for _, p := range c.Ports {
+				ports = append(ports, int(p.ContainerPort))
+			}
 		}
 	}
 
 	if containerFound == false {
-		return 0, fmt.Errorf("container '%s' not found", targetContainer)
+		return nil, fmt.Errorf("container '%s' not found", targetContainer)
```
michelleN: 👍

michelleN:
Thanks @radu-matei !

Issue #549, opened 2018-03-05
support packs with multiple files and chart folders


rawlingsj:
In our custom draft packs we include a Jenkinsfile and various other files in the root chart folder as well as extra folders that contain slightly different flavour of charts.  With the current `draft create` these extra files and folders are ignored and only the Dockerfile and a dir called "charts" is copied from the pack.

This is an example pack we have.  The folder 'preview' contains a modified helm chart that is used in CI to deploy the application.
```
packs
|__ java
     | Dockerfile
     | .dockerignore
     | Jenkinsfile
     |___ charts
     |___ preview
```

It would be super handy if draft would copy all files and folders in a pack rather than just the Dockerfile and charts folder.

We have a suggested solution to this which I'll followup in a PR but figured I ought to raise the issue to track why.

bacongobbler:
Hey! Since this is a duplicate of https://github.com/Azure/draft/issues/549 I'm going to go ahead and close this one so we can all work on one ticket. Please let me know if some use case is not addressed in that issue though and we can re-open this one. Thanks!

bacongobbler:
Hey! Since this is a duplicate of https://github.com/Azure/draft/issues/549 I'm going to go ahead and close this one so we can all work on one ticket. Please let me know if some use case is not addressed in that issue though and we can re-open this one. Thanks!

bacongobbler:
Just curious though...

> The folder 'preview' contains a modified helm chart that is used in CI to deploy the application.

What is different in the `preview` chart that cannot be handled in the regular chart?

rawlingsj:
Doh! I'd missed the duplicate issue and PR - my bad for not checking. 

> What is different in the preview chart that cannot be handled in the regular chart?

So in this scenario the preview folder contains a requirements.yaml which has a few extra apps that we use to create a preview environment for pull requests.  Here's a WIP example https://github.com/rawlingsj/draft/tree/master/packs/java

This is slightly different to when we install / upgrade apps in real environments as their requirements.yaml is stored in a separate git repo.

PR #453, opened 2017-12-16
copy all files from pack dir

bacongobbler:
closes #427

PR #550, opened 2018-03-05
Support packs with multiple files and multiple chart folders - fixes …

rawlingsj:
…#549

Rather than just copying a Dockerfile and folder called charts it would be handy if we can add extra files and folders to our draft packs.

We've found that we add extra files such as a Jenkinsfile (for CI/CD pipelines) and extra chart folders used for deploying into different environments.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ examples/**/.draftignore
 pkg/linguist/data/classifier
 pkg/linguist/data/linguist
 pkg/linguist/data/*.yml
+.idea/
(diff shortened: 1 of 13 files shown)


bacongobbler:
Hey @rawlingsj! Thank you for the PR, however this is a duplicate of https://github.com/Azure/draft/pull/453. There are some good ideas in this PR though, like recursively loading directories inside the pack as opposed to just the top-level files. Would you mind if we collaborated on that PR together via sending patches to https://github.com/bacongobbler/draft/tree/copy-pack-contents?

rawlingsj:
hi @bacongobbler! sure thing lemme take a look - thanks!

bacongobbler:
Awesome :)

AAAAAAAAND close.

PR #551, opened 2018-03-06
copy all files and folders from pack dir

rawlingsj:
Note this PR supersedes https://github.com/Azure/draft/pull/453 and combines commits from https://github.com/Azure/draft/pull/550 that copies all folders and well as all files.

I did attempt to patch the PR from @bacongobbler but as there was a rebase and quite a few conflicts it became very messy and hard to follow - this is that PR https://github.com/bacongobbler/draft/pull/2 which I couldn't get to work nicely so I closed.  Instead that same branch has no conflicts with Azure master and includes all the commits from #453 so am hoping things are clearer here, sorry not to reuse the same original PR but I hope it makes sense why.

__One thing to highlight__ - I reverted the [helm bump commit](https://github.com/Azure/draft/pull/453/commits/a55c640a9c85d9b3d289519dedf9a25be3a1179e) from #453, note that it's not being reverted from master, just the pull request.  I reverted it for a cpl of reasons:

- it was breaking CI, I couldn't get the build to compile 
- I wasn't sure if it was directly needed to copy the chart folders as the current code seems to work great without it
- I spent a couple of hours trying to figure out the godeps changes needed to get the helm bump to work but failed. I wondered, as the copying folders works without it the helm bump, is it better being added as a separate PR?


bacongobbler:
thanks @rawlingsj! I'll have a look at this.

.gitignore
```diff
@@ -12,3 +12,4 @@ examples/**/.draftignore
 pkg/linguist/data/classifier
 pkg/linguist/data/linguist
 pkg/linguist/data/*.yml
+.idea/
```
bacongobbler: Hey! Generally speaking we tend to keep IDE-specific stuff outside of .gitignore and in a (personal) global gitignore. e.g. on my system, all I have is:

```
><> git config --get core.excludesfile
~/.gitignore_global
><> cat ~/.gitignore_global 
.DS_Store
.vscode/
```

This allows us to keep the .gitignore specific to anything the project uses, and files that the OS or IDE generates up to the user.

With that in mind, would you mind removing this? :)

pkg/draft/pack/load.go
```diff
@@ -14,21 +16,31 @@ import (
...
+		return nil, fmt.Errorf("error reading %s: %s", topdir, err)
 	}
-
-	dockerfile := filepath.Join(topdir, DockerfileName)
-	pack.Dockerfile, err = ioutil.ReadFile(dockerfile)
-	if err != nil {
-		return nil, fmt.Errorf("error reading %s: %s", dockerfile, err)
+	for _, fInfo := range files {
+		if fInfo.IsDir() {
+			chart, err := chartutil.LoadDir(filepath.Join(topdir, fInfo.Name()))
```
bacongobbler: hmm... This makes the assumption that *every* directory in the pack is a chart. Do we really want to make that assumption?

.gitignore
rawlingsj: sure thing, done.

pkg/draft/pack/load.go
rawlingsj: Yeah I agree, it would be good to copy any directory rather than them having to be chart folders.  Do you have any ideas on how best to do that?

pkg/draft/pack/create_test.go
```diff
@@ -17,6 +18,13 @@ func TestCreateFrom(t *testing.T) {
 		t.Errorf("expected err to be nil, got %v", err)
 	}
 
+	// verify that canada.txt was copied over
+	if _, err := os.Stat(filepath.Join(tdir, "canada.txt")); err != nil {
+		if os.IsNotExist(err) {
+			t.Error("expected Canada to exist")
+		}
+	}
```
michelleN: 👍

Gopkg.lock
```diff
@@ -4,20 +4,28 @@
 [[projects]]
   name = "cloud.google.com/go"
   packages = ["compute/metadata"]
-  revision = "767c40d6a2e058483c25fa193e963a22da17236d"
-  version = "v0.18.0"
+  revision = "050b16d2314d5fc3d4c9a51e4cd5c7468e77f162"
+  version = "v0.17.0"
```
michelleN: how come we're rolling back off these pkg versions? Would you mind running a `dep ensure` if you haven't already?

README.md
```diff
@@ -39,8 +39,6 @@ This project has adopted the [Microsoft Open Source Code of Conduct](https://ope
 
 This software is covered under the MIT license. You can read the license [here][license].
 
-This software contains code from Heroku Buildpacks, which are also covered by the MIT license.
```
michelleN: How come we're removing this line?

pkg/draft/pack/load.go
michelleN: One way is to save the extra files as a map of file names and byte arrays(`files map[string][]byte`). We can save that on the `Pack` struct and then write those files in `pack.SaveDir`. If that seems like a sane suggestion for files, then we can extend that analogy further for copying directories and instead add a map of maps to the `Pack` struct: `directories map[string]files` with the string being the name of the directory.

Another way to implement this could be to leave the FromDir as is and in SaveDir, go back to the pack directory and loop through all the directories and copy into the current working directory leaving aside the Chart directory and the Dockerfile which have already been taken care of and are essential pieces in a `pack`

What do you think?

bacongobbler:
Hey @michelleN, those NOTICE file removements were from my PR in #453. I should probably split that out into another PR. Thank you for noticing that!

Gopkg.lock
rawlingsj: I think this was related to reverting the helm bump commit, not entirely sure TBH :) Once https://github.com/Azure/draft/pull/453 is merged however we should be able to rebase this PR and avoid any changes to Gopkg files.

pkg/draft/pack/load.go
rawlingsj: Thanks @michelleN yeah sounds good.  I'm a bit snowed under ATM so might wait for #453 to merge first and then this PR will be a bit cleaner if that's ok with you.

bacongobbler:
split out the notice files in #594. :)

radu-matei:
Can we do a rebase of this PR?
Is there anything blocking it?

Thanks!

squillace:
want want want

michelleN:
ping @rawlingsj

rawlingsj:
Yeah there still needs to be a bit of work on this to allow all folders be copied across and not just chart folders.  That aside I think we we aiming to have https://github.com/Azure/draft/pull/453 merged as this PR is based on it.

bacongobbler:
Hey @rawlingsj! We merged something similar in #741 so it looks like this is good to close now that we perform a deep copy of everything from the pack dir into the application directory. Let us know if there's issues with the new functionality in the canary release! Thanks for your contribution/feedback; we really appreciate it. :)

rawlingsj:
@bacongobbler awesome! I'll give it a whirl - thanks!

PR #558, opened 2018-03-12
use draft.toml's environment for `draft logs`

bacongobbler:
This allows us to remove the need for supplying the application name to `draft logs`.

### DIFF
--- a/cmd/draft/logs.go
+++ b/cmd/draft/logs.go
@@ -2,31 +2,35 @@ package main
 
 import (
 	"fmt"
+	"io"
+
 	"github.com/Azure/draft/pkg/draft"
+	"github.com/Azure/draft/pkg/draft/local"
 	"github.com/spf13/cobra"
 	"golang.org/x/net/context"
-	"io"
 )
 
 const logsDesc = `This command outputs logs from the draft server to help debug builds.`
 
 type logsCmd struct {
-	client   *draft.Client
-	out      io.Writer
-	appName  string
-	buildID  string
-	logLines int64
-	args     []string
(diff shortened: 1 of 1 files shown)


radu-matei:
There is also a [`GetLatestBuildID`](https://github.com/Azure/draft/pull/536/commits/a6c38a954395657e38bf62b283fd899598a14016) method that takes the app name as param - which would allow us to remove the need to pass the build ID if we want to get the logs regarding the latest build

bacongobbler:
That's a fantastic suggestion. Let's make that the next PR :)

bacongobbler:
Need to make the same change made in #554

radu-matei:
Sure, once #536 gets merged

cmd/draft/logs.go
```diff
@@ -51,13 +56,17 @@ func (l *logsCmd) complete(args []string) error {
...
 	}
-	l.appName = args[0]
-	l.buildID = args[1]
+	l.buildID = args[0]
 	return nil
 }
 
 func (l *logsCmd) run() error {
-	b, err := l.client.GetLogs(context.Background(), l.appName, l.buildID, draft.WithLogsLimit(l.logLines))
+	deployedApp, err := local.DeployedApplication(draftToml, l.runningEnvironment)
```
fibonacci1729: Maybe we should promote the runningEnvironment flag to the global set? It seems to be common to all subcommands.

bacongobbler:
closing in favour of #573 since the logging subsystem is being reworked there, and therefore this is being addressed there

PR #562, opened 2018-03-12
short-circuit image build/push step when image exists

### LINKED ISSUE #277: improve `draft up` docker image build performance
Right now when `draft up` is called twice in a row with no changes in-between, draft will re-build and re-push the same docker image to the registry, which is unnecessary. We should perform a smarter verification that the incoming docker image and chart are the same as the last release so we can tighten the dev loop in these cases.

bacongobbler:
closes #277

### DIFF
--- a/pkg/draft/server.go
+++ b/pkg/draft/server.go
@@ -159,9 +159,10 @@ func (s *Server) buildApp(ctx context.Context, req *rpc.UpRequest) <-chan *rpc.U
 	wg.Add(1)
 	go func() {
 		var (
-			buf = new(bytes.Buffer)
-			app *AppContext
-			err error
+			imageExists bool
+			buf         = new(bytes.Buffer)
+			app         *AppContext
+			err         error
 		)
 		defer func() {
 			s.finish(app, buf)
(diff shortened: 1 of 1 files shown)


bacongobbler:
Manually tested and working as intended:

```
><> draft up
Draft Up Started: 'example-go'
example-go: Building Docker Image: SUCCESS ⚓  (36.0435s)
example-go: Pushing Docker Image: SUCCESS ⚓  (16.0086s)
example-go: Releasing Application: SUCCESS ⚓  (0.1585s)
example-go: Build ID: 01C8E56ZTA15B6P39Z0MGMQR1Q
><> draft up
Draft Up Started: 'example-go'
example-go: Releasing Application: SUCCESS ⚓  (0.0394s)
example-go: Build ID: 01C8E57N8AVJQ5NDG2PYQJ8T89
```

Nopte the lack of a build and push step in the output :)

fibonacci1729:
What do you think of adding a `--force (-f)` flag here?

bacongobbler:
Hmm yeah that's a good idea. I'll add that in. Users should be able to have some control over the branch logic.

bacongobbler:
this should be revisited after #573, since we now skip the push step when a registry has not been configured.

bacongobbler:
closing for now. Let's see if the current performance is up to par first before adding new features on top of that. :)

Issue #272, opened 2017-08-18
implement `draft confg:{get,set,view}`


fibonacci1729:

radu-matei:
I think #573 closed this, right?

bacongobbler:
yep!

Issue #564, opened 2018-03-13
Safely passing "secrets" to docker build contexts


mumoshu:
A lengthy context about the secret management problem in docker: https://github.com/moby/moby/issues/13490

In nutshell, I need draft to support safely passing "secrets" from the local worktree to the remote build context.

A secret can be said to be safely passed when and only when the secret isn't exposed to malicious user or remained in any part of a resulting docker image.

A build secret is required when e.g. you're building a Ruby app depends on a private gem hosted in a private GitHub repository. In this specific case, you need a ssh key to download the private gem.

// Note that this is how `bundle install` or `gem`, the defacto dependency manager or the package manager for ruby works.

Non-safe cases include:

- Used docker's `--build-arg` to pass secrets: NG.
  - This results in the docker image has the secrets inside metadata can be seen by e.g. running `docker inspect`.
- Used `ADD` instruction to pass secrets: NG.
  - This results in the docker image contains the secret in one of its layers, can be queried by e.g. running `docker save $image | tar -x -C out && find out/ -type file | xargs -n 1 grep $secret`
  - You can indeed squash the image to a single layer so that ADDed and then `rm`ed secret disappear from the resulting docker image. But squashing results in docker-layer-caching useless, no incremental build. You end up re-building the whole docker image from scratch after each `draft up`...

--

Just for clarification, I'm not going to exploit the draft-built docker image itself for production-use.
I'll build the production one on my CI/CD system.

So, my concern here is solely about development, which I believe what draft stands for. I don't want to expose my precious secret even in dev env!

## Proposal

Today, the most general work-around for the problem seems to be utilizing a temporary tcp/http server last within the single docker build which holds/serves secrets to the build context.
You can basically utilize the server like `curl $endpoint/$secret -o ~/.ssh/id_rsa` && bundle install && rm ~/.ssh/id_rsa`, so that you can pass the secret without polluting the docker image metadata nor the  ADD instruction.

Existing docker-build alternatives like [habitus](https://github.com/cloud66/habitus) does support this feature.

Would it be good to improve draft to support it, too?

bacongobbler:
Hey @mumoshu! Nice to see you here. :)

I think you've summarized the issue and the landscape quite well.

[...]

Issue #432, opened 2017-11-09
Improve cycle time on Minikube by eliminating push step


benweissmann:
I'm working on setting up a development environment for my company based on minikube, and draft seems like a great fit -- except that the cycle time from saving a file to having it running on the cluster is longer than it'd like. Of the ~15 seconds it takes to update the service, about 8 seconds is pushing the image (which is a 19MB image).

For now, i've rolled my own upgrade path that does:

```
eval "$(minikube docker-env)"
docker build -t my-app .

kubectl delete pod -l app=my-app --grace-period=0 --force
```

and have the image tag in my helm package be "latest". This is much faster, but loses all the other nice benefits of draft.

Would you be open to adding an option to use a flow like this (probably with a `helm upgrade` or even a `kubectl set image`) when minikube is detected, rather than the default push-to-registry flow? I'm happy to help with implementation, but wanted to check whether this is something you'd consider merging. Pointers to where to look to implement this change would also be much appreciated.

bacongobbler:
Hi there!

Given Draft's current state, I'd be concerned about the user experience introducing this new workflow. How would this workflow work on different cloud providers like Azure or Google Kubernetes Engine, for example? We wouldn't want to introduce a workflow that works one way on one cloud provider that differs from the others. If we eliminate the push step, none of the kubelets (other than the one running the docker daemon we just built the image with) will be able to run the container.

That being said, we have been thinking about opening up the workflow to be scriptable using [brigade](http://brigade.sh/) as the backend for draft. That might open up new opportunities such as this. It's a much larger re-write than we may be able to tackle given the current roadmap for a 1.0 release, though.

In other words, I think this is a great use case to optimize for minikube, but we're not at the stage where we can start pulling out core functionality from Draft in certain situations yet. :)

bacongobbler:
There is also a related discussion in https://github.com/Azure/draft/issues/277 that discusses how we could potentially improve the 8 second image push time. That might be a quicker solution than what I proposed above.

jstrachan:
if a cluster has only 1 node then it makes sense as an optimisation to push directly to the docker daemon of the 1 node to speed up the developer round trip. Maybe some env var or something could enable the 'optimal single node' developer flow?

bacongobbler:
In certain instances, it’s not guaranteed that there will always be one node in the cluster. For example a lot of users trying out AKS are seeing the value in scaling down the cluster to a lower node count at night.

[...]

PR #573, opened 2018-03-15
remove draftd

### LINKED ISSUE #13: improve pkg cmd/draft/installer coverage
```
><> go test -cover -run . ./cmd/draft/installer
ok   github.com/deis/draft/cmd/draft/installer  0.006s  coverage: 0.0% of statements
```

### LINKED ISSUE #168: break up `draft init` into a plugin architecture

### LINKED ISSUE #178: write documentation around docker volume cleanup
Right now we don't clean up the docker volume. It would be nice to write some documentation on how users can do this themselves.

### LINKED ISSUE #213: authenticate against docker registry on `draft init`?
env: Azure Cloud Shell (Ubuntu 16.04)
Draft: 0.5.1
Helm: 2.5
Kubectl: 
Client Version: version.Info{Major:"1", Minor:"6", GitVersion:"v1.6.6", GitCommit:"7fa1c1756d8bc963f1a389f4a6937dc71f08ada2", GitTreeState:"clean", BuildDate:"2017-06-16T18:34:20Z",
GoVersion:"go1.7.6", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"6", GitVersion:"v1.6.6", GitCommit:"7fa1c1756d8bc963f1a389f4a6937dc71f08ada2", GitTreeState:"clean", BuildDate:"2017-06-16T18:21:54Z",
GoVersion:"go1.7.6", Compiler:"gc", Platform:"linux/amd64"}

Experience: 
draft --debug up gives:

```
 ---> b2e973c8c32e
Successfully built b2e973c8c32e
Successfully tagged bucket-microsoft.azurecr.io/bucket-microsoft/brawny-panda:dad28aeb9d351a00d2804b7dfda4285043789805
--> Pushing bucket-microsoft.azurecr.io/bucket-microsoft/brawny-panda:dad28aeb9d351a00d2804b7dfda4285043789805
The push refers to a repository [bucket-microsoft.azurecr.io/bucket-microsoft/brawny-panda]
ae943f533244: Preparing
2238e2393893: Preparing
ed7b605dc18c: Preparing
59d131791a5d: Preparing
be9cbeee35e4: Preparing
3b783483463c: Preparing
7f5c5258f618: Preparing
d9ac75d7d160: Preparing
19a735fbad89: Preparing
5fada12c27a7: Preparing
a420442c7891: Preparing
e4dafb7385d0: Preparing
e9c375282580: Preparing
dcd61d2ac531: Preparing
71ce2dc7f761: Preparing
0d960f1d4fba: Preparing
Error encountered streaming JSON response: no basic auth credentials
Watching local files for changes...
DEBU[0000] NOTIFY PATH: /home/ralph/draft/examples/dotnetcore/...
```

Expectation: `--debug` should have given some sort of hint where, precisely, this cred issue occurred, and which creds are mistaken or, _especially_, missing (if that's the case).

### LINKED ISSUE #247: use go-bindata to autogenerate the chart
Right now we have the chart that is installed by `draft init` embedded directly in the code as well as in `chart/`. This means that maintainers need to make changes in two different places, and it has gone out of sync many times. We should use `go-bindata` to autogenerate the go code based on the charts as we do with the draft packs in #108.

blocked by #236 as that will change the directory structure of the chart.

### LINKED ISSUE #285: Unable to push/pull from AWS ECR Container Registry
Draft currently has a few assumptions that mean that it won't work if you're using a private registry with AWS ECR.

The first is that there's an assumption of a an `org` prefix. So you end up with a final repo URL of `12345.dkr.ecr.us-east-1.amazonaws.com/orgname/imagename:tag`.

Due to ECRs bizarre set up, you don't usually have a repo prefix. But generally just a `12345.dkr.ecr.us-east-1.amazonaws.com/imagename:tag`

Also, the image repo actually has to exist before using it. So if you push into a repo image URL that doesn't exist, it *doesn't* create it by default (would be nice if it did).

And, finally, the username / password combination in AWS expires after 12 hours and needs to be refreshed.

docker pulls from an AWS instance node *generally* have IAM roles that allow you to pull from them, and there is a default system service which will refresh the tokens every 12 hours.

I'm not 100% sure how draft works, whether we could use these default permissions to make image pulling work out of the box.

Also, currently, `draft init` asks for username/password, but if you get the password details using `aws ecr get-login --no-include-email` you end up with a huge password token that won't actually paste into the cli.

I know it's super early days on draft, and it's a very exciting project, but I know a lot of the AWS crowd uses private registries, so without it - it's a non-starter.

Is there a workaround?

I considered editing `draftd` runtime parameters, and the secrets files. But without the 12 hour token refresh thing I'm not sure it would help much.

Keep up the amazing work!

### LINKED ISSUE #306: draft pre-flight check plugin baked into draft init
TBD: name for the plguin

Draft is built on other modular components to help developers create apps that will end up on Kubernetes. In some cases, you'll already have some dependencies. It'd be nice for `draft init` to not only check for these dependencies but also start or run installation steps if necessary. So for example, if minikube is not running, then start it. If Tiller is not installed, run the `helm init` command. We as a team have talked about creating a plugin that checks for the dependencies and configures them if they are not already running/installed. Then, we'd like to make that a "default" plugin that runs as part of the `draft init` command after the client side steps in the `draft init` logic.

### LINKED ISSUE #343: Draft up isn't upping
So I just checked out the latest version of draft 

```
kris-nova:terraformctl kris$ draft version
Client: &version.Version{SemVer:"v0.7.0", GitCommit:"0054537f5fe2561f8dc0fe3cca0fc5970f62562d", GitTreeState:"clean"}
Server: &version.Version{SemVer:"v0.7.0", GitCommit:"0054537f5fe2561f8dc0fe3cca0fc5970f62562d", GitTreeState:"clean"}
```

and whenever I run a `draft up` I am not getting a build ID and no pods are showing in my cluster.

```
kris-nova:terraformctl kris$ draft up --debug
DEBU[0000] Created tunnel using local port: '60626'
DEBU[0000] SERVER: "localhost:60626"
DEBU[0001] INCLUDES: [.]
DEBU[0001] EXCLUDES: [/go/src/github.com/kris-nova/terraformctl/chart]


Draft Up Started: 'kindled-rodent'
kindled-rodent: Build ID:
```

also I tried to `helm install` the chart that draft created just for good measure and am getting this error:

```
kris-nova:terraformctl kris$ helm install chart/go
Error: render error in "go/templates/ingress.yaml": template: go/templates/ingress.yaml:1:14: executing "go/templates/ingress.yaml" at <.Values.ingress.enab...>: can't evaluate field enabled in type interface {}
```

Any ideas?

### LINKED ISSUE #355: GCR support
I'd be good to support GCR auth. I'd be happy to hack something up, at least as PoC. It just needs a to fetch a token from metadata API.

In Flux we do this: [`github.com/weaveworks/flux/registry/gcp.go`](https://github.com/weaveworks/flux/blob/0e7ac0435cacdad42669ad1760b0a37f2715262f/registry/gcp.go)

Which is called from here: [`github.com/weaveworks/flux/registry/credentials.go`](https://github.com/weaveworks/flux/blob/0e7ac0435cacdad42669ad1760b0a37f2715262f/registry/credentials.go#L94)

Another simple option is to install `gcloud` in `dind` container, and running any `gcloud docke --` command populates the `/root/.dockercfg` with tokens.

There was a discussion in https://github.com/Azure/draft/issues/49, but died out with respect to GCR. Thoughts?

### LINKED ISSUE #356: [META] error: invalid reference format: repository name must be lowercase
There are several causes of this error:

```
time="2017-09-22T16:50:56.763305114Z" level=error msg="Handler for POST /v1.29/build returned error: invalid reference format: repository name must be lowercase"
```

One of them would be #209.

This also pops up when the registry add-on has not been enabled for minikube, which should hopefully be resolved with #213 or #306.

### LINKED ISSUE #404: Use Tiller/Helm ServiceAccount  (RBAC)
Currently is not possible to define `serviceAccount` for draft, so it will use default one in whatever namespace command is run. This cause troubles when running draft as part of CI/CD.

- using gitlab-runner in namespace `testing` for CI
- in last CI step this happens:

```
$ draft init --client-only
Creating /root/.draft 
Creating /root/.draft/plugins 
Creating /root/.draft/packs 
Creating pack gradle...
Creating pack python...
Creating pack go...
Creating pack php...
Creating pack ruby...
Creating pack javascript...
Creating pack java...
Creating pack csharp...
$DRAFT_HOME has been configured at /root/.draft.
Skipped installing Draft's server side component in Kubernetes due to 'client-only' flag having been set
Happy Sailing!
$ draft up
Error: User "system:serviceaccount:testing:default" cannot list pods in the namespace "kube-system". (get pods)
ERROR: Job failed: error executing remote command: command terminated with non-zero exit code: Error executing in Docker Container: 1
```

Even out of CI/CD system draft may need to deploy to different name-spaces (like Helm) 

I edited draftd deployment and added:

```
serviceAccount: tiller
serviceAccountName: tiller
```

but got same error

Is there any workaround?

### LINKED ISSUE #428: There's no way of refreshing a base image once it's pulled into draftd's dind
Once an image is pulled into draftd's dind, there's no way to refresh it.

In order to refresh an updated base image, I had to:

```
$ kubectl exec -it draftd-67bb886665-ntxz7 -c dind \
    docker pull docker-registry/draft/perl:onbuild
```

According to the following, the only build option passed is the resulting tag:

https://github.com/Azure/draft/blob/21cf4af77195423b5238adfdbfdd027da0de9695/pkg/draft/server.go#L147

An idea would be add to `draft up` a flag to indicate whether or not pulling the base image should be attempted by draftd.

### LINKED ISSUE #470: `draft init` to beta
Sets up plugins and client side packs, connects to (or sets up) container registry, installs or connects to tiller, starts draftd. Works against Minikube, Azure, AWS, GKE.

This is a meta issue to track `draft init`'s progress from alpha to beta. We'll finalize design and UX around `draft init` here.

Discussion topics/laundry list for `draft init`:

- [ ] #168
- [ ] #285
- [ ] #306
- [ ] #355

### LINKED ISSUE #479: draft init to allow container registry, user/pwd to be passed in as CLI arguments
if the docker registry, user and pwd cannot be deduced (e.g. for minikube) then the user is prompted for them on the CLI. it'd be nice to allow them to be passed on the CLI as arguments as well for easier scripting.

Happy to submit a PR if folks think its a reasonable thing to do

### LINKED ISSUE #482: Ksonnet
Will there be any relationship of Draft with Ksonnet for generating Helm charts?

### LINKED ISSUE #485: draft RBAC story
This is a meta issue for tracking the issue of users running Draft in RBAC-enabled Kubernetes clusters.

#358 raised this issue around how to work with draft in RBAC enabled Kubernetes clusters. @bacongobbler added documentation on this in #439 to address this. We still need to discuss and finalize what the experience around this should be and can use this issue to do so. Currently, we say that you'd need to set up `helm` with a `service-account` using `helm init --service-account=<service-account-name>` before running `draft init` and that may still stand. Feedback welcome.

### LINKED ISSUE #508: `draft init` to GA
This is a meta issue to track the progress of `draft init` from beta to stable.

### LINKED ISSUE #517: only enable --insecure-registry flags on minikube
Presently every draft cluster installed can talk insecurely to any container registry from within these CIDR ranges:

- 10.96.0.0/12
- 10.0.0.0/24

This is such that draftd on minikube can push to the registry add-on. We should be enabling this feature for only minikube clusters as other cloud providers (GKE, AKS) use their own container registry (GCR, ACR).

### LINKED ISSUE #546: [Wish] Ability to add insecure registry in draft setup
(Apologies if this is possible already and I have missed something in the docs)

Example: I am using the latest `minikube` (vm-driver=none) but the IP range my registry is assigned is `10.97.*.*` which falls outside of the ones currently defined.

It would be nice to have the ability to pass my own insecure registry ranges to `Draft` when I am installing it.  (Currently I have to go in and edit the `deployment.yaml` manually)

### LINKED ISSUE #548: Propagate Tiller TLS client cert to Draftd
If Tiller TLS is enabled, currently Draftd will not be able to communicate to it.

This issue tracks the work for adding a `tls-tiller-cert` flag to `draft init` to propagate the Tiller TLS client certificate to Draftd.

See: #544

bacongobbler:
This PR removes Draftd from the cluster and runs builds locally. This is probably one of the most significant changes ever made to Draft since the switch from websockets to gRPC, so hold onto your nickers!

ocean's eleven - the pinch

The most noteworthy information:

- draftd has been removed
- `draft init` serverside bootstrapping for draftd has been removed
- ingress configuration has been removed, since we have `draft connect` now and there's no need to set a `basedomain`
- ~~`draft create` now has a `--repository` flag, which is the name of the Docker registry to publish the image to (see the getting started guide in this PR for more info)~~ This has been replaced with `draft config set registry` and the `registry` flag in draft.toml
- Draft plugin's `useTunnel` now initiates a connection to Tiller instead of Draft
- re-named all `$DRAFT_NAMESPACE` and `--draft-namespace` flags to `$TILLER_NAMESPACE` and `--tiller-namespace`
- logs for `draft logs` are stored in `$DRAFT_HOME/logs`
- moved `pkg/storage/kube` to `pkg/storage/kube/configmap` (this would be easy to revert/pull into a separate PR, but figured I should do it while in there)
- configuration for talking to docker has been moved to `draft up`

enables use cases like #564 

closes #548 
closes #546 
closes #517 
closes #508 
closes #485 
closes #482 
closes #479 
closes #470 
closes #428 
closes #404 
closes #356
closes #355 
closes #343 
closes #306 
closes #285 
closes #247 
closes #213 
closes #178 
closes #168 
closes #13


bacongobbler:
TODO:

- [ ] fix `draft logs
- [ ] update the reference docs
- [ ] find and remove dead code

bacongobbler:
Removing the WIP :)

squillace:
On Ubuntu 16.04 LTS, and I suspect any env on which docker requires `sudo` by default -- obviously NOT MAC WITH DOCKER FOR MAC :-) -- we either need to notice that, and request permission on init, or document that you need to remove sudo from docker using the instructions at https://docs.docker.com/install/linux/linux-postinstall/#manage-docker-as-a-non-root-user.

squillace:
Not a blocker, as this can be documented first, and the proper fix settled upon.

squillace:
Builds fine; pushes to docker hub fine (azurecr never finishes pushing; doesn't do minikube by default, per above), and I get this on the example-go app in the PR branch. It appears to be looking for the ingress in the values field.

```
╰─± d up
Draft Up Started: 'example-go'
example-go: Building Docker Image: SUCCESS ⚓  (1.0003s)
example-go: Pushing Docker Image: SUCCESS ⚓  (7.0016s)
buildApp: release error: could not install release: rpc error: code = Unknown desc = render error in "go/templates/ingress.yaml": template: go/templates/ingress.yaml:1:14: executing "go/templates/ingress.yaml" at <.Values.ingress.enab...>: can't evaluate field enabled in type interface {}
example-go: Releasing Application: FAIL ❌  (0.0280s)
example-go: Build ID: 01C8SB78P6GAXE4C2RQM4KJGVW
Inspect the logs with `draft logs 01C8SB78P6GAXE4C2RQM4KJGVW`
```

squillace:
Hmmm, a different error on this one. I create a new version of the same app, and I got this. Pushing to ACR using `az acr login` worked, whereas using `docker login` to ACR did not. but: 

```
╰─± d up
Draft Up Started: 'squillace/example-go-fooishness'
squillace/example-go-fooishness: Building Docker Image: SUCCESS ⚓  (4.0006s)
squillace/example-go-fooishness: Pushing Docker Image: SUCCESS ⚓  (16.0030s)
buildApp: release error: could not upgrade release: rpc error: code = Unknown desc = invalid release name, must match regex ^(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])+$ and the length must not longer than 53
squillace/example-go-fooishness: Releasing Application: FAIL ❌  (0.0194s)
squillace/example-go-fooishness: Build ID: 01C8SCCPXWH59VFBQ3NHKEG4EM
Inspect the logs with `draft logs 01C8SCCPXWH59VFBQ3NHKEG4EM`
```

bacongobbler:
> Not a blocker, as this can be documented first, and the proper fix settled upon.

Good feedback. I'll document this.

bacongobbler:
> Not clear how to map the local image repository to the cluster's repository. Does one, for example, need to use something like https://hub.docker.com/r/faithlife/minikube-registry-proxy/? or do the mapping manually, like in https://abhishek-tiwari.com/local-development-environment-for-kubernetes-using-minikube/?

It needs to be one registry that is accessible both to the user (draft) and the cluster (minikube). This is the same model prior to this PR, however this was made slightly easier because the docker daemon was co-located with the registry in Minikube's use case. We may need to revisit the local `registry` plugin for minikube and how we ineract with it, but I don't think it's something we need to solve right out of the gate.

> It appears to be looking for the ingress in the values field.

I made changes to the default packs that removes the ingress rules, but `draft init` will install packs from master, not this branch. When testing, you'll need to override the packs using the following commands:

```
draft pack-repo remove github.com/Azure/draft
draft pack-repo add https://github.com/bacongobbler/draft --version remove-draftd
```

As for this error:

> this seems to be because I named the image `squillace/example-go-fooishness`, and it blew on deploying something with a / in it.

I think you made the `name` field in draft.toml `squillace/example-go-fooishness`, right? If so you need to make that `example-go-fooishness` and the `repository` field set to `squillace.azurecr.io`. Does that help make more sense?

bacongobbler:
I just added `draft config get|set|unset|list`, so this now also closes #272 to address initial comments about being unable to set a global OS-wide registry for all apps deployed by a single tenant. I also dropped the `--registry` flag from `draft create` to pronounce that setting a local `registry` field in draft.toml is intended to override what's in `$DRAFT_HOME/config.toml`.

I'll see about getting this sucker rebased.

bacongobbler:
Ooh, forgot to add back the ingress rules. one sec

radu-matei:
One thing I would like to have is streaming logs (at least for the Docker build)

bacongobbler:
Would it be all right if we opened a new ticket for that? I'm just trying to bring this PR down to the bare essentials so any future enhancements can be built on top of this :)

bacongobbler:
action item:

- eliminate push step when on minikube a la #432

bacongobbler:
oh, I also filed #585 as a proposal to store build information locally. For now we'll switch over to the configmap backend so commands like `draft history` will continue to work with this new architecture.

bacongobbler:
last action item: need to read in docker envvars so draft can use the docker daemon via `eval $(minikube docker-env)`.

bacongobbler:
All right, so for anyone who is interested in testing this PR:

## Testing draftd-less Draft

1. check out the code and compile `draft` from source and move it somewhere on your $PATH, as usual
1.  `rm -rf ~/.draft`
1. Run `draft init` to re-create ~/.draft
1. Now, remove the "default" packs and replace with this branch's packs: `draft pack-repo remove github.com/Azure/draft` and `draft pack-repo add $GOPATH/src/github.com/Azure/draft`

Now you're ready to set up the cluster of your choice.

### Minikube Setup

1. If you have the registry add-on enabled, feel free to disable it as it's no longer necessary for Draft to work. `minikube addons disable registry` should do it.
1. It's easiest to delete your existing minikube cluster and start from a clean slate: `minikube delete`
1. `minikube start`
1. Once the cluster is up, run `eval $(minikube docker-env)` to populate your environment with the appropriate environment variables so Draft can use Minikube's Docker daemon. NOTE: you should not need to install the `docker` daemon or the CLI on your host for this to work. Draft should just pick this up automatically.
1. Run `helm init` to start Tiller

### Azure Setup

Start by creating a kubernetes cluster and a container registry in a supported region such as 'eastus':

```
az group create -n remove-draftd -g eastus
az aks create -g remove-draftd -n myk8scluster
az acr create -g remove-draftd -n myacr --sku Standard
```

Because we need to build images locally and push those images into the container registry, we need to [install `docker`](https://docs.docker.com/install).

Once docker is installed, we need to make sure the current user can access and run `docker` commands. On Ubuntu, this can be done by adding the current user to the `docker` group:

```
usermod -aG docker $(whoami)
```

Log out, then log back in as yourself for it to take effect.

Now we can log into the container registry:

```
az acr login -n myacr -g remove-draftd
```

And finally, tell Draft to push images to this registry:

```
draft config set registry myacr.azurecr.io
```

## Deploy an App

Now with the cluster of our choice set up, we can finally deploy an application.

```
$ cd examples/example-go
$ draft create
--> Draft detected Go (100.000000%)
--> Ready to sail
$ draft up
```

If anything isn't working according to the steps provided here, please let me know.

pkg/builder/builder.go
```diff
@@ -0,0 +1,540 @@
...
+		}()
+		if app, err = newAppContext(b, bctx, buf); err != nil {
+			fmt.Printf("buildApp: error creating app context: %v\n", err)
+			return
+		}
+		if err := b.buildImg(ctx, app, ch); err != nil {
+			fmt.Printf("buildApp: buildImg error: %v\n", err)
+			return
+		}
+		if app.ctx.Env.Registry != "" {
```
radu-matei: If we are not on Minikube and the registry is not set, this will skip the pushing step and will try to release the chart without the image being pushed to a registry.

The Draft output is `SUCCESS`, but it actually silently fails as it cannot find the image.

squillace: oops

radu-matei: Sample output:

```
Draft Up Started: 'example-go'
example-go: Building Docker Image: SUCCESS ⚓  (5.0139s)
example-go: Releasing Application: SUCCESS ⚓  (0.9168s)
example-go: Build ID: 01C950XBFNB14Y7CKQ90BHRRN9
Inspect the logs with `draft logs 01C950XBFNB14Y7CKQ90BHRRN9`
```

bacongobbler: This is silently intentional. We have no way of detecting whether or not we are actually running against a non-minikube cluster. I've made a note of it in both the getting started guide as well as the reference docs, though I don't really see a way around this.

Any suggestions?

bacongobbler: The hack I implemented long ago in `draft init` mightt work here (relying on the kubernetes config's current-context name to be "minikube", but it was more or less a crutch and probably won't hold up forever.

bacongobbler: Would it be fair to inform the user that no registry was set before starting the build?

radu-matei: The least we can do I think is outputting that the push step is skipped because there is no registry set in the config.

It would at least inform the user why.

Another thing that I have in mind - can we detect if it's Minikube by the master URL when using the Kubernetes client? It's surely localhost when Minikube, right?

squillace: I don't like the "minikube" magic string solution, as I now have five different minikube profiles for different configurations, as you can imagine. I prefer to a) inform the person that no registry was set, so if it's not running against minikube it will skip the push activity and report success and yet fail; and b) doc the heck out of it.

However, another option might be to have the default be to push, but have the --skip-push flag such that when you know you're local, do this to skip push? Dunno. Whatchya think.

radu-matei: Oh, GitHub refresh...

bacongobbler: @squillace I considered the --skip-push flag, but that means minikube users now need to run `draft up --skip-push` all the time. It's a little more cumbersome for an "optional" feature flag, so I wanted to make the default experience sane.

I've made a change in 75fd33f so it'll display a warning, as well as a way to disable that warning if you know what you're doing. Sounds good?

squillace: Let's try it for a bit, and if it's an issue we can identify a better ux. at this point, it's really a "did you remember to specify a registry if you're working against the cloud?" problem. It's not really a draft problem.

squillace:
On Ubuntu, rebuild for last commit, delete .draft, draft init, minikube: 

```
╰─± ../../bin/draft up
WARNING: no registry has been set, therefore Draft will not push to a container registry.
This can be fixed by running

        $ draft config set registry docker.io/myusername
Hint: this warning can be disabled by running

        $ draft config set disable-push-warning 1
Draft Up Started: 'example-go'
example-go: Building Docker Image -buildApp: buildImg error: Post http://192.168.99.100:2376/v1.23/build?buildargs=null&cachefrom=null&cgroupparent=&cpuperiod=0&cpuquota=0&cpusetcpus=&cpusetmems=&cpushares=0&dockerfile=&labels=null&memory=0&memswap=0&networkmode=&rm=0&shmsize=0&t=example-go%3A3a8762f94ec2e318dd55af9b08c9e53d345bd7c0&target=&ulimits=null: net/http: HTTP/1.x transport connection broken: malformed HTTP response "\x15\x03\x01\x00\x02\x02\x16".
* Are you trying to connect to a TLS-enabled daemon without TLS?
example-go: Building Docker Image: FAIL ❌  (1.0002s)
example-go: Build ID: 01C95ZESW75PVZEEMQ1CCBMPG3
Inspect the logs with `draft logs 01C95ZESW75PVZEEMQ1CCBMPG3`
```

I'll restart minikube.

squillace:
stopping, minikube delete, minikube start, rm .draft, eval step, draft init, helm init, same error.

commit:
commit e1a391394a426c0c6ebacafe8495605f2b7707fe
Author: Matthew Fisher <matt.fisher@microsoft.com>
Date:   Wed Mar 21 13:11:42 2018 -0700

    s/step/warning

bacongobbler:
Hmm, I hit that same error. I thought I fixed that earlier. Lemme try and debug the situation. Thanks for bringing that up!

bacongobbler:
ok, I think I fixed that up now. Feel free to give 'er another shot! :)

bacongobbler:
that was a doozie. Thanks to everyone involved for the feedback and reviews!

squillace:
ACK, @bacongobbler. If yoga boy got past that issue, then you don't have to wait for me, but I'll rebuild and retest.

bacongobbler:
Yoga boy did indeed get past this, haha :)

> you beat me to the punch with your last fix to the PR. But with that in, I have to express: https://gph.is/1r1DGY7 :boom:

squillace:
ETA for merge? want canary image built instantly as possible to start demo work with binaries. without devving fulltime, I lose track of branches/builds I did locally. :(bad PM, bad PM).

squillace:
me love me some demonless.

PR #576, opened 2018-03-17
Initial prep work for custom tags for drafted docker images

### LINKED ISSUE #541: `draft up` should tag and release a ":latest" tag
Draft makes it very hard to reuse the deployed image in other pods because the tag is neither provided during `draft up` nor tagged as latest.
I suggest to add the ability to tag the built image so that other pods may be able to use it.

I'm currently using pachyderm which has it's own way to deploy pods to Kubernetes so I can't use a helm chart to deploy them as of yet.

thedrow:
This PR adds the latest tag to the docker image. This is useful because latest is a floating tag and will allow users to reuse the image outside the context of a `draft up` invocation if needed.
This PR also makes image names a plural which is needed to fix #541.

Ref #541 but does not fix it yet.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -1092,6 +1092,6 @@
 [solve-meta]
   analyzer-name = "dep"
   analyzer-version = 1
-  inputs-digest = "09d18ea0a7471f02dc11146d4b00bf98f305601fc7ec6f28c01e052a8c46f90d"
+  inputs-digest = "e598e395392d2ba1f7a23155fda00ea3ed10ed370e7b05b5bb2f98c2ef1c5219"
   solver-name = "gps-cdcl"
   solver-version = 1
--- a/pkg/draft/context.go
+++ b/pkg/draft/context.go
@@ -26,7 +26,7 @@ type AppContext struct {
 	req  *rpc.UpRequest
 	buf  *bytes.Buffer
 	tag  string
-	img  string
+	imgs []string
 	out  io.Writer
 	id   string
 	vals chartutil.Values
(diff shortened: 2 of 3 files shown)


thedrow:
Currently this is what I need to continue working. I'd love to help out more but I'd need some guidance through the code base to do so.
I'm also not sure this is the right fix. So I'd appreciate a code review.

bacongobbler:
I just tested this and found the image push step going from a 16 second push (at best) to a 32 second push. I mentioned that there's some limitations with Docker and pushing duplicate image layers to the registry in https://github.com/Azure/draft/issues/432#issuecomment-355611968, so while normally this would be a great feature, it looks like this particular fix will hit performance in a significant way.

Given that we're trying to make performance improvements for the base case, I'd prefer not to go ahead with this change as is. That being said, if you can figure out a way to make the image push on identical layers much quicker, I'd be all for this!

thedrow:
That's very surprising. If you try to do same from the docker CLI you don't get such a performance degradation.
The fix for #541 involves a lot more code but it could make pushing additional image tags optional.
if you could guide me through the necessary changes, I'd be happy to fix the issue entirely.
First of all, we need to be able to pass multiple tags on the command line. I'm also guessing that UpReqest should be modified. What else am I missing?

bacongobbler:
Hey @thedrow! Thank you so much for the PR. Given this PR significantly affects performance for the base case and is enabled by default, we cannot accept it as is. With that in mind, I'm going to close this PR, but this does not mean we are saying "no" to this feature; just not in it's current form. :)

If this was somehow made as an opt-in feature, that would be fine!

The infrastructure was changed quite significantly in #573, so this PR will also require a significant amount of work to be rebased.

If you feel like making this an opt-in feature and have it ready for another review, please feel free to either re-open this PR or open a new one.

Thanks!

PR #580, opened 2018-03-19
 Add override-ports flag and use dashes instead of underscores in draft.toml

radu-matei:
This PR:

- switches underscores for dashes for `draft.toml`, as discussed in Slack
- adds an `override-ports` flag in `draft.toml`

Thanks!


bacongobbler:
nice. thanks!

PR #583, opened 2018-03-20
Wait for pod to be ready before connecting

### LINKED ISSUE #329: draft connect should wait for pods to come up before establishing connection

radu-matei:
This PR fixes #329 

- can someone with an available Minikube test it? At some point @michelleN did some testing and got some timeout errors - I haven't seen any issues on AKS so far

### DIFF
--- a/pkg/kube/podutil/podutil.go
+++ b/pkg/kube/podutil/podutil.go
@@ -1,8 +1,17 @@
-// package podutil exists for functions that exist in k8s.io/kubernetes but not in k8s.io/client-go. Everything here should be contributed upstream.
+// package podutil exists for functions that exist in k8s.io/kubernetes but not in k8s.io/client-go. Most of the things here should be contributed upstream.
 
 package podutil
 
-import "k8s.io/api/core/v1"
+import (
+	"fmt"
+	"time"
+
+	"k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
+)
 
 // IsPodReady returns true if a pod is ready; false otherwise.
 func IsPodReady(pod *v1.Pod) bool {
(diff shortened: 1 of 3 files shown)


pkg/kube/podutil/podutil.go
```diff
@@ -35,3 +42,30 @@ func GetPodCondition(status *v1.PodStatus, conditionType v1.PodConditionType) (i
 	}
 	return -1, nil
 }
+
+// GetPod waits for a pod with the specified label to be ready, then returns it
+// if no pod is ready, it checks every second until a pod is ready
+func GetPod(namespace, labelKey, labelValue string, clientset kubernetes.Interface) *v1.Pod {
```
bacongobbler: Would it be better to accept a label selector instead? e.g. https://github.com/kubernetes/helm/blob/48e703997016f3edeb4f0b90e6cfdb3456ce3db0/pkg/helm/portforwarder/portforwarder.go#L58-L59

radu-matei: Yes, it makes sense as will also add a label with the build ID later on.

bacongobbler:
One thing I noticed was the lack of a timeout. Does that exist in this PR? If we're waiting for a pod to come ready, is it possible that this function will never exit?

radu-matei:
Yeah, totally makes sense - how many seconds do you think is reasonable to wait for pods to be ready?
I'm thinking somewhere around 30 seconds?

bacongobbler:
Helm has a timeout of 5 minutes which is pretty reasonable.

PR #594, opened 2018-03-22
remove Heroku notices

bacongobbler:
We've no longer relied on Heroku buildpacks for quite some time.


bacongobbler:
nearly a year ago, Draft relied on Heroku's [`bin/detect`](https://github.com/Azure/draft/blob/v0.5.0/packs/python/detect) scripts from their buildpacks to perform language detection. Those scripts were removed in #239 when we introduced linguist. These licenses got accidentally left behind in that PR so we should be safe to remove it :)

PR #599, opened 2018-03-22
Fix root command's pre run not being called

radu-matei:
Because of https://github.com/spf13/cobra/issues/216, since `draft up` has its own  `PersistentPreRun`, the `rootCmd` `PersistentPreRunE` from `draft.go` is no longer called - which means (among others) that the global config is not loaded.

This adds a call to the  the `rootCmd` `PersistentPreRunE`.

cc @bacongobbler


bacongobbler:
Nice find. LGTM

PR #600, opened 2018-03-23
Auto connect for draft up

radu-matei:
This PR adds the `draft up --auto-connect` flag and entry in manifest.

In order to ignore existing running pods with the previous version, this PR also injects the build ID as label on the pods and takes it into consideration when connecting.

### DIFF
--- a/cmd/draft/connect.go
+++ b/cmd/draft/connect.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"os"
 	"os/signal"
 	"syscall"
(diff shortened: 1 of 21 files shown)


radu-matei:
Because we added the build ID as label, when doing `draft up` on an existing deployment, you get:

```
/buildApp: release error: could not upgrade release: rpc error: code = Unknown desc = Deployment.apps "example-go-go" is invalid: spec.template.metadata.labels: Invalid value: map[string]string{"app":"example-go-go", "buildID":"01C9HK23XFPTVY009XTER91D6M", "draft":"example-go"}: `selector` does not match template `labels`
```

Thoughts to avoid this?

michelleN:
oh dang one more conflict @radu-matei -- probably from all the PR merges to master today 😅

PR #605, opened 2018-03-26
chore(Gopkg.lock): format Gopkg.lock

michelleN:
This doesn't change any versions, just simplifies
format of Gopkg.lock so we don't see changes after
running `make bootstrap`


michelleN:
oh yeah you're right @radu-matei! I upgraded dep and didn't have that format issue anymore. Thanks!

PR #606, opened 2018-03-26
Add docs about draft connect

radu-matei:


docs/getting-started.md
```diff
@@ -90,24 +92,24 @@ example-python: Build ID: 01BSY5R8J45QHG9D3B17PAXMGN
 Now that the application has been deployed, we can connect to our app.
 
 ```shell
-$ draft connect
-Connect to python:8080 on localhost:39551
+$ draft connect -p 8081:8080
```
michelleN: small nit. Since this is a getting started guide, we may not need to put the `-p` flag example here since that is more of an advanced use case and this doc is for beginners. In addition, since the `override-ports` field is shown in the example `draft.toml`, the `-p` flag scenario may be confusing. What do you think?

docs/getting-started.md
```diff
@@ -90,24 +92,24 @@ example-python: Build ID: 01BSY5R8J45QHG9D3B17PAXMGN
...
 ```shell
-$ curl localhost:55139
+$ curl localhost:8081
 Hello, World!
 ```
 
 Once you're done playing with this app, cancel out of the `draft connect` session using CTRL+C.
 
+> Note that you can use the flag `draft up --auto-connect` in order to have the application automatically connect once the deployment is done, and you can customize the local ports for the application either using the `-p` flag or through the `override-ports` field in `draft.toml`. More info in [dep-007.md][dep007]
+
```
michelleN: This made me think that I could specify the ports on the `draft up` command since it was all in the same sentence. I think it might be more clear if we first specify that we can customize ports using the `-p` flag and override-ports field and then mention in a separate sentence that we can also use ` draft up --auto-connect` for convenience for the same behavior. Let me know if that makes sense.

docs/getting-started.md
bacongobbler: I agree. We should really strive for simplification where possible. `-p` isn't necessary in the base case, so I think it should be omitted from the Getting Started guides.

radu-matei:
Addressed initial comments 😄

PR #617, opened 2018-03-27
Avoid panic for draft connect and logs when there is no build in logs

radu-matei:
How to test this:

- delete logs in `~/.draft/logs`
- try running `draft connect` or `draft logs`

Instead of a panic, you should now get an error.


radu-matei:
Thanks 😄 

See #618

PR #629, opened 2018-03-28
re-instate injecting the image pull secret into the namespace

bacongobbler:
This enables use cases like deploying to an AKS cluster with a private third party registry such as quay.io. This is the same behaviour as draftd used to work in previous releases. I removed it in #573 because I thought all the use cases we were seeing involved a level of trust between the cluster and the registry, but that is not always the case.

This feature is only enabled if the user opts into setting a registry via `draft config set registry`. It performs a lookup in `~/.docker/config` to fetch the config, then injects that into the cluster via an image pull secret.

### DIFF
--- a/docs/install-advanced.md
+++ b/docs/install-advanced.md
@@ -38,7 +38,7 @@ We'll also need to log into the cluster to push images from our local docker dae
 $ az acr login -n myregistry -g myresourcegroup
 ```
 
-
+NOTE: Before deploying the chart to the cluster, Draft will inject a registry auth secret into the destination namespace so the image can be pulled into the cluster. You do not need to add a container registry secret into your chart; Draft will do this for you.
 
 ## Running Tiller with RBAC enabled
 
--- a/pkg/builder/builder.go
+++ b/pkg/builder/builder.go
@@ -3,6 +3,7 @@ package builder
 import (
 	"bytes"
 	"crypto/sha256"
+	"encoding/json"
 	"fmt"
 	"io"
 	"io/ioutil"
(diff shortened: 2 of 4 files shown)


squillace:
ACK. me likeee. it's not the requirement that is a problem; it's the UX - this makes it much smoother. I think we should eventually migrate toward making "docker" installation locally a thing of the past, but not yet, not yet. :-)

squillace:
question: does this "happen" only after you do `draft config set registry <url>`, or does it occur at some other time? I'm assuming after, and don't have time to read the code because unicorns and I'm your PM. Does that work?

robertstarmer:
That's how I read the code, only if you define a registry does it collect the key (which you should have already set, and lives in ~/.docker/config.json).

robertstarmer:
I'm basing that on:
https://github.com/Azure/draft/pull/629/files#diff-7d55094d2095596359f9af68227cd7cfR469

bacongobbler:
Yep! However it doesn't inject the image pull secret into the cluster until we are ready to release the application. That is to say, it occurs during `draft up`, not `draft config set`.

The reasoning behind that is that we've already confirmed that we can push the image to the container registry with the auth credentials, so we've confirmed that the auth credentials work from the docker engine -> the registry.

docs/install-advanced.md
```diff
@@ -25,7 +25,7 @@ We'll also need to log into the cluster to push images from our local docker dae
 $ az acr login -n myregistry -g myresourcegroup
 ```
 
-NOTE: Kubernetes distributions like Azure Container Service are automatically authorized to pull from container registries in the same resource group. If this is not the case, then you'll need to [add a container registry secret to your chart](https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry) to pull the private image.
+NOTE: Before deploying the chart to the cluster, Draft will inject a registry auth secret into the destination namespace so the image can be pulled into the cluster. You do not need to add a container registry secret into your chart; Draft will do this for you.
```
michelleN: 👍

bacongobbler:
ooh good catch. Perhaps we need a way to figure out how to pull the auth secret from the osx keychain. I'll have to play around with this some more.

bacongobbler:
fixed and added tests around converting a docker registry auth token to a kubernetes registry auth secret

squillace:
@bacongobbler I'd love to get this merged if you and @michelleN think the fix is good now.

PR #634, opened 2018-03-30
Add --dry-run flag to draft connect

radu-matei:
This flag is used by external tools to know when a pod of the application is running.


radu-matei:
This PR has the correct flag name, `--dry-run` and is disabled by default

PR #638, opened 2018-04-02
write logs to file instead of buffer

fibonacci1729:


radu-matei:
Would it make sense to have a flag on `draft up` that will stream the logs directly to `stdout`?

Without it, having streaming logs from an app external to Draft would need to have the log file (you must know the latest build ID + Draft home) or repeatedly execute `draft logs` and print the differences.

Thoughts?

fibonacci1729:
I agree. Mainly this was address from the context of #638. Given the UX we currently have, streaming directly to stdout is involved.

To stream logs, with this PR, the user can run:
```
$ tail -f $(draft home)/logs/<build-id>
```

PR #643, opened 2018-04-05
add tail support for draft logs

fibonacci1729:


cmd/draft/logs.go
```diff
@@ -47,8 +49,8 @@ func newLogsCmd(out io.Writer) *cobra.Command {
 	}
 
 	f := cmd.Flags()
-	f.Int64Var(&lc.logLines, "tail", 100, "lines of recent log lines to display")
-
+	f.BoolVar(&lc.tail, "tail", false, "tail the logs file as it's being written")
+	f.UintVar(&lc.line, "line", 20, "line location to tail from (offset from end of file)")
```
bacongobbler: out of curiosity, can one set `--line -1` to tail from the very beginning of the file?

bacongobbler:
just needs the tests to pass and we're good to go :)

cmd/draft/logs.go
fibonacci1729: Nope: `Error: invalid argument "-1" for "--line" flag: strconv.ParseUint: parsing "-1": invalid syntax`

bacongobbler:
So I think the issue has to do with transitive go dependencies introduced with github.com/hpcloud/tail. What we can try to fix this is alias `gopkg.in/fsnotify.v1` with v1.4.7 of https://github.com/fsnotify/fsnotify. That should get it working.

Something like this inside Gopkg.toml might work (note the use of `override` instead of `constraint`):

```
[[override]]
  name = "gopkg.in/fsnotify.v1"
  source = "github.com/fsnotify/fsnotify"
  version = "1.7.4"
```

Can you try that and see what happens?

Issue #627, opened 2018-03-28
Draft pack-repo canary release is installed with the stable releases


bacongobbler:
I think there's a bug in how we install the pack-repo plugin on `draft init`. If a user installs  a stable release such as v0.12.0, It should install a known stable release rather than the canary release.

bacongobbler:
fixed in #657

PR #657, opened 2018-04-09
inject plugin version into the plugin environment

bacongobbler:
This way, install hooks invoked via `draft plugin install` know which version of the plugin to install.

relates to #627. After this, we need to give draft-pack-repo knowledge of this envvar.


bacongobbler:
thanks for the review @michelleN!

PR #661, opened 2018-04-10
fix(client): avoid cloning map

### LINKED ISSUE #223: draft up with `set` fields does not work
Given the following draft.toml:

```
[environments]

    [environments.development]
    name = "myapp"
    set = ["ingress.enabled=true"]
```

That value is not set in the resulting chart. I think this appears because `.vals()` [completely ignores the environment's set values](https://github.com/Azure/draft/blob/82eda3d1a9230cd125f683a25d376b7b35f7fd92/cmd/draft/up.go#L96).

technosophos:
Remove `AsMap()` so that map doesn't get cloned.

Closes #223

### DIFF
--- a/pkg/builder/builder.go
+++ b/pkg/builder/builder.go
@@ -227,7 +227,7 @@ func loadArchive(ctx *Context) (err error) {
 func loadValues(ctx *Context) error {
 	var vals = make(chartutil.Values)
 	for _, val := range ctx.Env.Values {
-		if err := strvals.ParseInto(val, vals.AsMap()); err != nil {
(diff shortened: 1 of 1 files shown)


technosophos:
On closer inspection, what appears to have been happening is that `AsMap()` was returning an empty `map[string]interface{}` any time an empty values item was passed in. Calling AsMap is actually unnecessary, since Go's type system correctly infers. So that just side-steps the problem.

michelleN:
Should this go into 0.13.0 or 0.13.1? cc/ @bacongobbler

PR #670, opened 2018-04-13
feat(cmd/draft): override default plugins

michelleN:
pass in --config/-f flag on draft init to override
default plugins that get installed


bacongobbler:
Yessssss I see insane value in this feature. :heart:

michelleN:
To test this PR, create a file (we'll call is plugins.yaml):
```
[[plugin]]
name = "env"
url = "https://github.com/bacongobbler/draft-env"
version = "1.0.0"
```
Then run `draft init --config plugins.yaml`

cmd/draft/init.go
```diff
@@ -45,15 +49,21 @@ func newInitCmd(out io.Writer, in io.Reader) *cobra.Command {
...
 	f.BoolVar(&i.dryRun, "dry-run", false, "go through all the steps without actually installing anything. Mostly used along with --debug for debugging purposes.")
+	f.StringVarP(&i.configFile, "config", "f", "", "specify default plugins and pack repositories in a TOML file")
 
 	return cmd
 }
 
 // runInit initializes local config and installs Draft to Kubernetes Cluster
 func (i *initCmd) run() error {
 
+	plugins, repos, err := parseConfigFile(i.configFile)
```
bacongobbler: is it still possible to run `draft init` without this file present?

michelleN: oh goodness. I'll fix that first before merging. good catch!!

PR #674, opened 2018-04-15
Support for custom image tags

radu-matei:
This PR adds support for custom image tags.
To test, build this branch and add the following config in `draft.toml`:

```
custom-tags = ["latest", "your-custom-tag"]
```

The tags `latest` and `your-custom-tag` will be pushed to the container registry, as well as the auto-generated SHA tag that is currently pushed by Draft.

### DIFF
--- a/docs/reference/dep-006.md
+++ b/docs/reference/dep-006.md
@@ -27,12 +27,15 @@ The format of this file is as follows:
     watch-delay = 1
     override-ports = ["8080:8080"]
     auto-connect = false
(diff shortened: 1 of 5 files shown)


bacongobbler:
lgtm as an initial approach. In the future, it might be good to look into server-side image tagging via something like https://github.com/genuinetools/reg/issues/88 so we can support this workflow across any image building tool.

bacongobbler:
also, docs :)

PR #675, opened 2018-04-16
WIP: pull container builders into a separate package

bacongobbler:
This pulls the container builder implementation out into a separate package, opening up use cases to provide alternative workflows to build and push containers.

Some notes to consider:

- container builders only need to understand how to build and push images. They are given the kitchen sink provided by `draft up`
- pkg `builder` still understands how to deploy the image; that workflow doesn't change.
- each container builder still handles state through pkg/builder

### DIFF
--- a/cmd/draft/up.go
+++ b/cmd/draft/up.go
@@ -7,6 +7,7 @@ import (
 	"path/filepath"
 
 	"github.com/Azure/draft/pkg/builder"
+	"github.com/Azure/draft/pkg/builder/docker"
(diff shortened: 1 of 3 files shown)


technosophos:
So if I understand this correctly, it's the first step toward abstracting away the builder implementation, right? It looks like a good move to me.

bacongobbler:
yep!

bacongobbler:
I've still got a few design choices to re-assess here. I'll remove the WIP label once it's ready

PR #701, opened 2018-04-25
enable detaching from draft connect

fibonacci1729:
Running `draft connect --detach` will reexecute draft with `draft connect --export`. `--export` is a hidden flag and is meant to only be used by draft if invoked with `--detach`. `--export` instructs draft to dump the environment information for connection. Container logs are lost in the detached state.

Draft does not cleanup the detached process. The onus of cleanup is on the detacher, because draft is not a process manager. Re-running `draft connect --detach` will establish a new background draft tunnel.

Usecase:
```
/> eval $(draft connect --detach)
/> curl $EXAMPLE_GO_GO_SERVICE_HOST:$EXAMPLE_GO_GO_SERVICE_PORT
Hello World, I'm Golang!
```

### DIFF
--- a/cmd/draft/connect.go
+++ b/cmd/draft/connect.go
@@ -6,8 +6,10 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
+	"os/exec"
 	"os/signal"
 	"path/filepath"
+	"strings"
 	"syscall"
 	"time"
 
@@ -25,8 +27,9 @@ const (
 var (
 	targetContainer string
 	overridePorts   []string
-
-	dryRun bool
+	dryRun          bool
+	detach          bool
+	export          bool
 )
 
 type connectCmd struct {
@@ -45,6 +48,9 @@ func newConnectCmd(out io.Writer) *cobra.Command {
 		Short: "connect to your application locally",
 		Long:  connectDesc,
 		RunE: func(cmd *cobra.Command, args []string) error {
+			if detach {
+				return cc.detach()
+			}
 			return cc.run(runningEnvironment)
 		},
 	}
@@ -55,6 +61,9 @@ func newConnectCmd(out io.Writer) *cobra.Command {
 	f.StringVarP(&targetContainer, "container", "c", "", "name of the container to connect to")
 	f.StringSliceVarP(&overridePorts, "override-port", "p", []string{}, "specify a local port to connect to, in the form <local>:<remote>")
 	f.BoolVarP(&dryRun, "dry-run", "", false, "when this flag is used, draft connect will wait to find a ready pod then exit")
+	f.BoolVarP(&detach, "detach", "", false, "detach from the connection while preserving the tunnel")
+	f.BoolVarP(&export, "export", "", false, "export connection environment in detached state (hidden)")
+	f.MarkHidden("export")
 
 	return cmd
 }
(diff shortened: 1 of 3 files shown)


cmd/draft/connect.go
```diff
@@ -96,41 +105,85 @@ func (cn *connectCmd) run(runningEnvironment string) (err error) {
...
-		os.Exit(0)
+		close(done)
 	}()
 
-	for {
-		time.Sleep(5 * time.Minute)
-		fmt.Fprintf(cn.out, connectionMessage)
+	if export {
+		for env, val := range exportEnv {
+			fmt.Fprintf(os.Stdout, "export %s=%s\n", env, val)
```
bacongobbler: Can we make this cross-platform compatible?

The good news is that we can just re-apply the same logic `gofish tank` uses here :)

https://github.com/fishworks/gofish/blob/master/cmd/gofish/tank.go
https://github.com/fishworks/gofish/blob/master/cmd/gofish/tank_unix.go
https://github.com/fishworks/gofish/blob/master/cmd/gofish/tank_windows.go

cmd/draft/connect.go
```diff
@@ -96,41 +105,85 @@ func (cn *connectCmd) run(runningEnvironment string) (err error) {
...
+	cmd.Env = os.Environ()
+	if err := cmd.Start(); err != nil {
+		return err
+	}
+	b, err := ioutil.ReadAll(stdout)
+	if err != nil {
+		return err
 	}
+	cmd.Process.Release()
+	fmt.Println(string(b))
```
bacongobbler: since stdout from `draft connect --export` already prints an extra newline, can we just call `fmt.Print()` instead, or trim whitespace via `fmt.Println(strings.TrimSpace(string(b)))`? The output right now looks like this:

```
><> draft connect --detach
export EXAMPLE_GO_GO_SERVICE_HOST=localhost
export EXAMPLE_GO_GO_SERVICE_PORT=35505

><>
```

cmd/draft/connect.go
```diff
@@ -96,41 +105,85 @@ func (cn *connectCmd) run(runningEnvironment string) (err error) {
...
-	}
-
-	for _, cc := range connection.ContainerConnections {
-		readCloser, err := connection.RequestLogStream(deployedApp.Namespace, cc.ContainerName, cn.logLines)
-		if err != nil {
-			return err
+			if export {
+				var (
+					application = strings.Replace(strings.ToUpper(deployedApp.Name), "-", "_", -1)
+					container   = strings.ToUpper(cc.ContainerName)
```
bacongobbler: container names contain hyphens as well. Can we replace those with underscores too?

```
><> draft connect --detach
export EXAMPLE_GO_GO-CRAZY_SERVICE_HOST=localhost
export EXAMPLE_GO_GO-CRAZY_SERVICE_PORT=40743
```

cmd/draft/connect.go
```diff
@@ -96,41 +105,85 @@ func (cn *connectCmd) run(runningEnvironment string) (err error) {
...
+				fmt.Fprintf(cn.out, connectionMessage)
+			case <-done:
+				return nil
+			}
+		}
+	}
+}
+
+func (cn *connectCmd) detach() error {
+	cmd := exec.Command("draft", "connect", "--export")
```
bacongobbler: how about this?

```
cmd := exec.Command(os.Args[0], "connect", "--export")
```

That way we get the right path to *this* particular invocation of Draft.

cmd/draft/connect.go
bacongobbler: Also, what do you think about writing a few comments at the end of the string to make it a bit more user-friendly? e.g. something like this:

```
fmt.Printf("%s\n# # Run this command to configure your shell:\n# eval $(%s connect --detach)\n\n# To kill the underlying process:\n# sudo kill %d\n", strings.TrimSpace(string(b)), os.Args[0], cmd.Process.Pid)
```

cmd/draft/connect.go
```diff
@@ -55,6 +61,9 @@ func newConnectCmd(out io.Writer) *cobra.Command {
 	f.StringVarP(&targetContainer, "container", "c", "", "name of the container to connect to")
 	f.StringSliceVarP(&overridePorts, "override-port", "p", []string{}, "specify a local port to connect to, in the form <local>:<remote>")
 	f.BoolVarP(&dryRun, "dry-run", "", false, "when this flag is used, draft connect will wait to find a ready pod then exit")
+	f.BoolVarP(&detach, "detach", "", false, "detach from the connection while preserving the tunnel")
+	f.BoolVarP(&export, "export", "", false, "export connection environment in detached state (hidden)")
+	f.MarkHidden("export")
```
bacongobbler: nice, TIL

cmd/draft/connect.go
bacongobbler: revised edit:

```
fmt.Printf("%s\n# Run this command to configure your shell:\n# eval $(%s connect --detach)\n", strings.TrimSpace(string(b)), os.Args[0])
```

bacongobbler: and for Windows, something equivalent to https://github.com/fishworks/gofish/blob/master/cmd/gofish/tank_windows.go#L21 would be +:100:

fibonacci1729: Given how this works, It wouldn't make much sense to instruct them to run the eval command on export because re-running will spawn a new detached draft process. It would be nice to be able to re-export the same environment on every detach if one already exists but there's no way to do this just yet because detached draft's are not tracked between invocations.

bacongobbler:
sounds good. @squillace in the interest of time, would it be okay to follow up in another PR to document this?

squillace:
absolutely. don't care about long elaboration, just something so that searching if you need it finds it. Shouldn't take a second.

cmd/draft/connect.go
```diff
@@ -96,41 +105,92 @@ func (cn *connectCmd) run(runningEnvironment string) (err error) {
...
+			}
+		}
+	}
+}
+
+func (cn *connectCmd) detach() error {
+	args := []string{"connect", "--export"}
+	if len(overridePorts) > 0 {
+		args = append(args, "-p")
+		args = append(args, overridePorts...)
```
bacongobbler: I think the proper syntax is

```
draft connect -p 8080:8080 -p 8080:8081
```

So you need to iterate through each overridePort and append a `-p` flag rather than just once.

bacongobbler: so this could be simplified to

```
for i := range overridePorts {
	args = append(args, "-p")
	args = append(args, overridePorts[i])
}
```

PR #702, opened 2018-04-26
docs for draft connect detach mode

fibonacci1729:


docs/reference/dep-007.md
```diff
@@ -158,3 +158,22 @@ Connect to multi-port:80 on localhost:9090
...
+Using the `--detach` flag in `draft connect` will spawn a new Draft process that exports 
+the tunnel connection environment to standard out. All flags, excluding `--detach` will be
+propagated to the detached Draft process. A hidden `--export` flag is appended to this set
+of flags which instructs Draft to export the connection environment. As a result, container
+logs are lost if Draft is connected in detach mode.
+
+Continuing with the examples above:
+```
+$ eval $(draft connect -p 8081:8080 -c go --detach)
+$ curl localhost:8081
```
bacongobbler: did this mean to be `$ curl $EXAMPLE_GO_GO_SERVICE_HOST:$EXAMPLE_GO_GO_SERVICE_PORT` instead?

bacongobbler: oh wait no, this is right. never mind!

bacongobbler: we should probably remove the `eval` then since it's kinda useless with this example

Issue #715, opened 2018-04-30
`draft init` fails on Windows with space in username 0.14.1


dave-hillier:
Attempting to run `draft init`  fails with the following output:

```
PS C:\Users\Dave Hillier> draft init
Installing default plugins...
Installation of default plugins complete
Installing default pack repositories...
Error: exec: "C:\\Users\\Dave": file does not exist
```
And the version info:
```
PS C:\Users\Dave Hillier> draft version
&version.Version{SemVer:"v0.14.1", GitCommit:"fdc29c553a45600ac4f795f3485d4bb9a80c7862", GitTreeState:"clean"}
```

radu-matei:
Closing as it was addressed in #717 
Let us know if you still experience any more issues related to this.

Thanks!

PR #717, opened 2018-05-02
fix init on windows

birdayz:
Refer to #715 

While reproducing this, i hit three problems:

1) Running the plugin - PrepareCommand tries to determine the executable by splitting with " " - in case of whitespace in path it will compute a wrong path for the executable. See my changes - i work around it by expanding the environment var after splitting, and not before. This is not super clean, but works.

2) At least for me, when the hook get.ps1 is called, on my Win10 VM, the window closes immediatly. This is due to the WindowStyle Hidden flag when calling powershell for the hook. It still works correctly, but it's strange if your cli window just disappears (and makes debugging a nightmare)

3) Downloading the default plugin pack - The install hook of the default plugin repo has to be adjusted, because powershell requires a specific notation with an "&" if you require support for Commands that include whitespaces. I made a separate PR for this on the draft-pack-repo: https://github.com/draftcreate/draft-pack-repo/pull/27
Note that the version for the draft-pack has to be updated afterwards, if this gets merged.

I'm very happy if somebody tests these changes, i'm a linux user and not very comfortable with powershell. It works on my Win10 VM.

### DIFF
--- a/cmd/draft/plugin_windows.go
+++ b/cmd/draft/plugin_windows.go
@@ -22,7 +22,7 @@ func runHook(p *plugin.Plugin, event string) error {
 		return nil
 	}
 
-	prog := exec.Command("powershell.exe", "-ExecutionPolicy", "Bypass", "-NoLogo", "-NonInteractive", "-NoProfile", "-WindowStyle", "Hidden", "-Command", hook)
+	prog := exec.Command("powershell.exe", "-ExecutionPolicy", "Bypass", "-NoLogo", "-NonInteractive", "-NoProfile", "-Command", hook)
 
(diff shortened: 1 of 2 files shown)


bacongobbler:
Thanks @birdayz! I'm away from my Windows box while at KubeCon and will be on vacation until the 13th, but if someone else gets to test this that'd be awesome. Thank you for working on a fix!

bacongobbler:
Hey, thanks for the ping. I've just cross-compiled this and will boot into Windows to test this out. Thanks again, @birdayz! SO sorry about the delay.

bacongobbler:
just tested and confirmed this works for me on Windows. Thanks so much @birdayz for the fix!

PR #721, opened 2018-05-04
Suppress colorization when it would result in escape sequence noise

itowlson:
When running `draft up` on Windows under `cmd` or `powershell`, the colourisation escape sequences are not interpreted but just printed to the screen, resulting in confusing noise around the Draft progress messages.  This PR suppresses colourisation on Windows, making messages more readable.

It would be good to detect Windows shells which are capable of handling escape sequences, e.g. `cmder` or `hyper` - this is a first step that addresses what I think is going to be the most common case for Draft on Windows.  It can be refined if required by improving the `cmdline.consoleSupportsColor()` function over time.

### DIFF
--- a/pkg/cmdline/cmdline.go
+++ b/pkg/cmdline/cmdline.go
@@ -2,6 +2,7 @@ package cmdline
 
 import (
 	"fmt"
+	"runtime"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


bacongobbler:
Thanks Ivan!

PR #731, opened 2018-05-05
Docs: Dont allow pushing upstream

teaguecole:
I noticed that in the draft/contributing/hacking.md, there was nothing about preventing someone from making the mistake of pushing changes upstream to master.  I edited the doc to include the simple git command to prevent that from happening. #newbiecontributor

### DIFF
--- a/docs/contributing/hacking.md
+++ b/docs/contributing/hacking.md
@@ -53,10 +53,11 @@ $ cd draft
 ```
 
 Add the conventional [upstream][] `git` remote in order to fetch changes from Draft's main master
-branch and to create pull requests:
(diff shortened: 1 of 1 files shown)


birdayz:
i don't understand it.. most people are not allowed to push to upstream master anyway, so what's the point?

teaguecole:
If that is the case I can close this

bacongobbler:
Yeah, this is unnecessary since we've got the master branch protected, so this is not required. IIRC Kubernetes has also got their master branch protected so that documentation is likely out of date. Branch protection in GitHub was released for general use after Kubernetes was introduced.

Thanks for checking though! :)

screenshot_20180511-165401

PR #741, opened 2018-05-11
ref(*): deep copy from draft pack to current dir

### LINKED ISSUE #743: draft create should copy all files from a draft pack over to directory
Currently `draft create` copies only top level files from a draft pack over to the application directory. If there are nested files in directories, they should be copied over too.

michelleN:
* draft create currently only copies top level files
* This PR allows draft to copy directories and nested files
over to the current working directory as well

resolves #743


bacongobbler:
supercedes #551

Issue #742, opened 2018-05-11
revamp docs


michelleN:
It's hard for newcomers to know that DEPs are the place to find documentation around Draft features. DEPs should be used to propose features and discuss architecture/rationale. They're not great for telling users how or when to use Draft/packs/plugins/other Draft features.

We should look to put more user-centric docs under the `docs/` dir like [Helm](https://github.com/kubernetes/helm/tree/master/docs) does.

PR #744, opened 2018-05-11
feat(docs): add quickstart guide & README

michelleN:
taking a stab at re-vamping docs #742

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -1,0 +1,39 @@
+# Quickstart Guide
+
+This guide covers how you can quickly get started using Draft
+
+## Prerequisites
+
+The following prerequisites are required for successful use of Draft.
+
+1. A Kubernetes cluster and the `kubectl` CLI tool
+2. Installation of Helm on your Kubernetes cluster
+3. Installing and configuring Draft on your laptop
+
+### Install Kubernetes or have access to a cluster
+- If you don't already have a running Kubernetes cluster, consider downloading and installing [Minikube][] which can help you get a Kubernetes cluster running on your local machine
+- You can download and configure the `kubectl` binary using the directions [here](https://kubernetes.io/docs/tasks/tools/install-kubectl/)
+
+### Install and configure Helm
+- Helm is a package manager for Kubernetes. It allows you to install and manage packages of Kubernetes manifests, more commonly known as "charts", on your Kubernetes cluster.
+- Download the Helm binary using [Homebrew](https://brew.sh/) via `brew install kubernetes-helm` or the [official releases page](https://github.com/kubernetes/helm/releases).
+- Once you have Helm on your machine and a running Kubernetes cluster, run the following command:
+```console
+$ helm init
+```
+
+### Install and configure Draft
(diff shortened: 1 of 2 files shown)


docs/quickstart.md
```diff
@@ -0,0 +1,36 @@
...
+## Prerequisites
+
+The following prerequisites are required for successful use of Draft.
+
+1. A Kubernetes cluster and the `kubectl` CLI tool
+2. Installation of Helm on your Kubernetes cluster
+3. Installing and configuring Draft on your laptop
+
+### Install Kubernetes or have access to a cluster
+- If you don't already have a running Kubernetes cluster, consider downloading and installing [Minikube][] which can help you get a Kubernetes cluster running on your local machine
```
radu-matei: Should we consider also adding Docker for Mac/Windows here, as the Kubernetes feature will probaly get out of the beta channel around DockerCon?

I will test all the steps here with Docker for Mac and maybe add a PR after we know everything works.

docs/quickstart.md
```diff
@@ -0,0 +1,36 @@
...
+
+### Install and configure Helm
+- Helm is a package manager for Kubernetes. It allows you to install and manage packages of Kubernetes manifests, more commonly known as "charts", on your Kubernetes cluster.
+- Download the Helm binary using [Homebrew](https://brew.sh/) via `brew install kubernetes-helm` or the [official releases page](https://github.com/kubernetes/helm/releases).
+- Once you have Helm on your machine and a running Kubernetes cluster, run the following command:
+```console
+$ helm init
+```
+
+### Install and configure Draft
```
radu-matei: @bacongobbler - should we add Fish instructions for Draft?

docs/quickstart.md
```diff
@@ -0,0 +1,36 @@
...
+```console
+$ brew tap azure/draft
+$ brew install draft
+```
+- Set up Draft on your machine by running:
+```console
+$ draft init
+```
+
+Congratulations! You're all set! Check out the [Getting Started](getting-started.md) page to see how to use Draft with a sample application.
```
radu-matei: Before linking to the Getting Started guide, which starts with a `draft create` and `draft up`, should we first point to an installation guide? Or add the info here on how to config the registry / use the correct Docker context for Minikube?

docs/quickstart.md
bacongobbler: Yes; I'd love to.

palma21: Maybe chocolatey as well? Especially for windows

bacongobbler: Theres no chocolatey package available for Draft. We're tracking that in another issue, but once that's available we can add that to the list as well.

bacongobbler: I will note however that GoFish works on Mac/Linux/Windows so it's something Windows users can try out.

@palma21 for reference: https://gofish.sh

palma21: Yes I also use gofish but just stumbled on https://chocolatey.org/packages/draft hence the comment

docs/quickstart.md
michelleN: oh good catch! I added that part in based on your feedback. Thanks

Issue #748, opened 2018-05-12
Java example fails to run in container / pod


mlgibbons:
The Java example fails as it is generating a JAR file with dependencies (Uber JAR) and the Java pack's Docker file uses the first JAR file that it find i.e. the JAR file without dependencies. The result is that when the container runs a ClassNotFound exception is raised due to the Spark web frameworks classes not being in the JAR file.

Issue #727, opened 2018-05-05
Command in the Dockerfile for a Java application


arun-gupta:
The command in the Dockerfile generated for a Java project is:

```
CMD ["/bin/bash", "-c", "find -type f -name '*.jar' | xargs java -jar"]
```

Depending upon what kind of Java application is being used, the `target` directory could have a large number of JAR files. For example, for the sample application at https://github.com/arun-gupta/microservices-greeting, the list of generated JARs is:

```
./target/greeting/WEB-INF/lib/aopalliance-repackaged-2.5.0-b42.jar
./target/greeting/WEB-INF/lib/asm-6.0.jar
./target/greeting/WEB-INF/lib/asm-commons-6.0.jar
./target/greeting/WEB-INF/lib/asm-tree-6.0.jar
./target/greeting/WEB-INF/lib/aws-java-sdk-core-1.11.321.jar
./target/greeting/WEB-INF/lib/aws-lambda-java-core-1.2.0.jar
./target/greeting/WEB-INF/lib/aws-serverless-java-container-core-1.0.jar
./target/greeting/WEB-INF/lib/aws-serverless-java-container-jersey-1.0.jar
./target/greeting/WEB-INF/lib/aws-xray-recorder-sdk-core-1.2.1.jar
./target/greeting/WEB-INF/lib/bootstrap-2018.2.0.jar
./target/greeting/WEB-INF/lib/commons-codec-1.9.jar
./target/greeting/WEB-INF/lib/commons-fileupload-1.3.3.jar
./target/greeting/WEB-INF/lib/commons-io-2.2.jar
./target/greeting/WEB-INF/lib/commons-logging-1.1.3.jar
./target/greeting/WEB-INF/lib/config-api-1.2.0.jar
./target/greeting/WEB-INF/lib/config-api-runtime-1.2.0.jar
./target/greeting/WEB-INF/lib/container-2018.2.0.jar
./target/greeting/WEB-INF/lib/ee-2018.2.0.jar
./target/greeting/WEB-INF/lib/elytron-2018.2.0.jar
./target/greeting/WEB-INF/lib/hk2-api-2.5.0-b42.jar
./target/greeting/WEB-INF/lib/hk2-locator-2.5.0-b42.jar
./target/greeting/WEB-INF/lib/hk2-utils-2.5.0-b42.jar
./target/greeting/WEB-INF/lib/httpasyncclient-4.1.1.jar
[...]

PR #749, opened 2018-05-12
Changed to use JAR file with dependencies

mlgibbons:
Fix to address issue: https://github.com/Azure/draft/issues/748

### DIFF
--- a/packs/java/Dockerfile
+++ b/packs/java/Dockerfile
@@ -9,4 +9,4 @@ EXPOSE 4567
 COPY --from=BUILD /usr/src/app/target /opt/target
 WORKDIR /opt/target
 
-CMD ["/bin/bash", "-c", "find -type f -name '*.jar' | xargs java -jar"]
+CMD ["/bin/bash", "-c", "find -type f -name '*-with-dependencies.jar' | xargs java -jar"]


michelleN:
cc/ @arun-gupta in case he has anything to add here

arun-gupta:
I suspect this is related to https://github.com/Azure/draft/issues/727. Java pack's Dockerfile tries to search for all the files in a directory and invoke them one by one. I think more intelligence is needed in the Java pack to figure out the exact JAR file that should be invoked instead of a generic `find` command.

arun-gupta:
And the file names may not necessarily be `*-with-dependencies.jar`. For example, WildFly Swarm (now Thorntail) generates the JAR file name as `*-swarm.jar`.

squillace:
I get it. We will have some Java heads look hard at it and circle around here. Thanks for the comments!

mlgibbons:
Hi. There are various ways to address the lack of genericity in the Java buildpack and some discussion is clearly warranted. Fixing the fact that the current pack fails is IMHO a pressing problem as it makes draft look bad for a first time user downloading and trying the Java example; as I did. May I suggest accepting the pull request and I will raise an issue to fix the limitation i.e. "Modified Java buildpack handles only Ubers jars". How does that sounds?

michelleN:
That makes sense @mlgibbons. Thank you for contributing!

Let's continue to track the conversation on how to make this a more generic experience in #727

PR #752, opened 2018-05-14
add appveyor.yml

bacongobbler:

PR #753, opened 2018-05-14
bump helm to v2.9.0

bacongobbler:


bacongobbler:
the appveyor check can be ignored for now until #752; Appveyor still runs a build even if no appveyor.yml is present.

Gopkg.toml
```diff
@@ -34,15 +34,39 @@
...
   revision = "6633656539c1639d9d78127b7d47c622b5d7b6dc"
 
+[[override]]
+  name = "github.com/russross/blackfriday"
+  version = "v1.4-2-g300106c"
+
 [[override]]
   name = "google.golang.org/grpc"
-  version = "=v1.2.1"
+  version = "1.7.2"
```
michelleN: do we need to override the grpc version or even have this anymore since we don't have a server side component?

bacongobbler: Good point. I had to add this to fix up some version conflicts but I'll see if I can just remove it

bacongobbler: Actually it looks like google.golang.org/grpc is still needed in Draft for `draft delete`: https://github.com/Azure/draft/blob/974526fbf0e7d7ece543519effe172d078c64d11/cmd/draft/delete.go#L100

Helm bumped up from GRPC 1.2.1 to 1.7.2 in 2.8 or 2.9 that had some compatibility breakers, and because kubernetes or Helm don't use dep we still need to define what version to use or else 1.2.1 will be satisfied since that will satisfy dep's SAT when analyzing Draft's usage of GRPC in the codebase.

I did a quick look through the rest of Gopkg.toml and the rest all look like packages we still need/require for Draft to work. Thanks for checking this though! :)

bacongobbler:
With that said, it's time to ship it :sunglasses:

PR #755, opened 2018-05-14
feat(docs): add examples and tutorials page

### LINKED ISSUE #754: where should we put more complex examples
I built an example of a rails app with all of the draft bits [here](https://github.com/michelleN/todo-app)

It's a bit larger than our examples though, so I'm thinking we may not want to actually merge this into the draft repo. Instead, should we have a doc that points to these kinds of examples in the docs?

michelleN:
resolves #754

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -7,7 +7,7 @@ This guide covers how you can quickly get started using Draft
 The following prerequisites are required for successful use of Draft.
 
 1. A Kubernetes cluster and the `kubectl` CLI tool
-2. Installation of Helm on your Kubernetes cluster
+2. Installation of [Helm](https://github.com/Kubernetes/helm) on your Kubernetes cluster
 3. Installing and configuring Draft on your laptop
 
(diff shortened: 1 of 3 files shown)


docs/quickstart.md
```diff
@@ -7,7 +7,7 @@ This guide covers how you can quickly get started using Draft
 The following prerequisites are required for successful use of Draft.
 
 1. A Kubernetes cluster and the `kubectl` CLI tool
-2. Installation of Helm on your Kubernetes cluster
+2. Installation of [Helm](https://github.com/Kubernetes/helm) on your Kubernetes cluster
```
bacongobbler: should kubernetes be lowercase?

michelleN: I don't actually know. Let me look around!

michelleN: It looks like in other places in our docs and comments, it is capitalized.

michelleN: I'll leave as is to preserve consistency but open to changing in the future if necessary!

PR #772, opened 2018-05-20
Implement image build arguments container builder

### LINKED ISSUE #663: draft up - Support for docker build options
My organization is in the process of evaluating draft and so far we are extremely happy with it. 

I ran into an issue today where one of our developers asked me about authenticating to a private node module registry during the docker build phase of `draft up`. 

The official node [documentation](https://docs.npmjs.com/private-modules/docker-and-private-modules) recommends passing the npm token as a `--build-arg`, which doesn't appear to be supported by `draft up`. 

We have a couple of ugly workarounds that we've been toying with but It would be great to be able to leverage the full range of docker build options (not just `--build-arg`) as there are likely many other use cases out there waiting to be discovered.

radu-matei:
This PR adds build arguments for the Docker builder.

To test it: 

- add the following in `draft.toml`:

```
    image-build-args = { BUILD_TIME_ARG = "build-time-argument-value",  HTTP_PROXY = "http://my-proxy" }
```

- use the arguments in your Dockerfile:

```
ARG BUILD_TIME_ARG
RUN echo $BUILD_TIME_ARG
ARG HTTP_PROXY
RUN echo $HTTP_PROXY
```

Here are the [best practices for Docker build arguments](https://docs.docker.com/engine/reference/commandline/build/#set-build-time-variables---build-arg), linked in the dep here.

closes #663

### DIFF
--- a/pkg/builder/azure/builder.go
+++ b/pkg/builder/azure/builder.go
@@ -76,11 +76,25 @@ func (b *Builder) Build(ctx context.Context, app *builder.AppContext, out chan<-
 			imageNames = append(imageNames, fmt.Sprintf("%s:%s", app.Ctx.Env.Name, imageNameParts[len(imageNameParts)-1]))
 		}
 
+		var args []containerregistry.BuildArgument
+
+		// TODO: once the API includes this as default, remove it
+		buildArgType := "DockerBuildArgument"
+		for k := range app.Ctx.Env.ImageBuildArgs {
+			name := k
+			value := app.Ctx.Env.ImageBuildArgs[k]
+			arg := containerregistry.BuildArgument{
+				Type:  &buildArgType,
+				Name:  &name,
+				Value: &value,
+			}
+			args = append(args, arg)
+		}
+
 		req := containerregistry.QuickBuildRequest{
 			ImageNames:     to.StringSlicePtr(imageNames),
 			SourceLocation: sourceUploadDefinition.RelativePath,
-			// TODO: make this configurable with https://github.com/Azure/draft/issues/663
-			BuildArguments: nil,
+			BuildArguments: &args,
 			IsPushEnabled:  to.BoolPtr(true),
 			Timeout:        to.Int32Ptr(600),
 			Platform: &containerregistry.PlatformProperties{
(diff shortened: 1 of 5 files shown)


radu-matei:
Added support for build arguments for ACR Build

fibonacci1729:
Great! Looks good to me!

bacongobbler:
Looks like the CLA bot was having some issues earlier this morning, but the rest of CI passed so let's go ahead and merge this. Thanks so much @radu-matei!

pkg/builder/azure/builder.go
```diff
@@ -76,11 +76,25 @@ func (b *Builder) Build(ctx context.Context, app *builder.AppContext, out chan<-
 			imageNames = append(imageNames, fmt.Sprintf("%s:%s", app.Ctx.Env.Name, imageNameParts[len(imageNameParts)-1]))
 		}
 
+		var args []containerregistry.BuildArgument
+
+		// TODO: once the API includes this as default, remove it
+		buildArgType := "DockerBuildArgument"
+		for k := range app.Ctx.Env.ImageBuildArgs {
+			name := k
+			value := app.Ctx.Env.ImageBuildArgs[k]
```
bacongobbler: I think this can be removed if we use the form `for k, value := range app.Ctx.Env.ImageBuildArgs {`. Would that work?

pkg/builder/docker/builder.go
```diff
@@ -32,9 +32,16 @@ func (b *Builder) Build(ctx context.Context, app *builder.AppContext, out chan<-
 	msgc := make(chan string)
 	errc := make(chan error)
 	go func() {
+		args := make(map[string]*string)
+		for k := range app.Ctx.Env.ImageBuildArgs {
+			v := app.Ctx.Env.ImageBuildArgs[k]
```
bacongobbler: I think this can be removed if we use the form `for k, v := range app.Ctx.Env.ImageBuildArgs {`. Would that work?

bacongobbler:
bumping to 0.16.0 so we can go ahead and get a release out today. :)

pkg/builder/azure/builder.go
radu-matei: Not necessarily - since both Docker and ACR expect a pointer to a string, if we pass `&v` from `for k, v := range ...`, then it doesn't take the pointer to the actual loop value, but to the loop variable.

The same is valid for your other comment in this PR - the alternative would be to use `to.StringPtr(value)` - using the `autorest` method - it's just a matter of preference that I wanted to do it without an external package.

bacongobbler: aw dang yeah I missed that. Thanks for pointing that out! :)

mtsmfm:
When will you cut new release?

PR #778, opened 2018-05-24
add erlang pack and example

tsloughter:


packs/erlang/Dockerfile
```diff
@@ -0,0 +1,40 @@
+FROM erlang:20.3.6-alpine as builder
+
+RUN apk add --update tar curl git bash make libc-dev gcc g++ && \
+    rm -rf /var/cache/apk/*
+
+RUN set -xe \
+    && curl -fSL -o rebar3 "https://s3.amazonaws.com/rebar3-nightly/rebar3" \
```
michelleN: hey @tsloughter - what is rebar3? is this an erlang package? are erlang packages normally stored in s3?

bacongobbler: Looks like a build tool for Erlang apps

Would it be better to use a stable release instead of the nightlies? https://github.com/erlang/rebar3/releases

tsloughter: Thanks, good catch. I've updated it to use the tagged release. I upload those to `s3://rebar3/rebar3` after publishing them on github, so that s3 link is always the latest stable release.

michelleN: awesome thank you

michelleN:
For some reason, it says `unverified` next to the latest commit. I've never seen that before. Have you @bacongobbler ? Not sure if it's something we should pay attention to.

tsloughter:
Seems like something I messed up with my signing of commits, should be safe to ignore.

bacongobbler:
Yea [according to your keybase profile](https://keybase.io/tsloughter) the signing key should really be C702BA4186F93DCF. I just went though this myself so I'm happy to share my git config if you'd like but it seems safe to ignore. :

tsloughter:
Thanks, but the rest of my commits today on other repos are verified, so I think my config is fine. Must have been something last night or with my rebasing..

PR #798, opened 2018-05-30
Fix wrong Dockerfile path when running draft up from another directory

### LINKED ISSUE #720: Relative Dockerfile path fails to build
I'm currently unable to build a docker image if it's in a relative path on Windows... 😢 

```
PS > draft up
...
app-dev: Building Docker Image: FAIL ❌ (1.0000s)

PS > draft logs 01CCKHJMBRJZGQ7G7PAKJSBP0W
2018/05/03 10:42:43 error while building: Error response from daemon: Cannot locate specified Dockerfile: ./api/Dockerfile

PS > docker build ./api/
# works fine
Successfully built 8f8f4524e2f7

PS > draft version
&version.Version{SemVer:"v0.14.0", GitCommit:"e7a92f6b2f05a23f4cd9cc15fc24eb98d628e5d7", GitTreeState:"clean"}
```

radu-matei:
To test this:

- change the name of your Dockerfile in the directory and change that name in `draft.toml` - for example, using `example-go` from this repo:

```
[environments]
  [environments.development]
    name = "testing-golang"
    namespace = "default"
    custom-tags = ["latest", "canary"]
    dockerfile = "Dockerfile.golang"
```

- run: `draft up examples/example-go` and everything should work as intented

Anyone around who can test on Windows before I do?

closes #720 

Thanks, @bacongobbler, for quickly identifying the issue!

### DIFF
--- a/pkg/builder/builder.go
+++ b/pkg/builder/builder.go
@@ -264,7 +264,9 @@ func loadValues(ctx *Context) error {
 }
 
 func archiveSrc(ctx *Context) error {
-	contextDir, relDockerfile, err := build.GetContextFromLocalDir(ctx.AppDir, ctx.Env.Dockerfile)
+
+	dockerfilePath := filepath.Join(ctx.AppDir, ctx.Env.Dockerfile)
+	contextDir, relDockerfile, err := build.GetContextFromLocalDir(ctx.AppDir, dockerfilePath)
 	if err != nil {
 		return fmt.Errorf("unable to prepare docker context: %s", err)
 	}


bacongobbler:
Looking closely at `getDockerfileRelPath` from docker/cli, it does seem like it'll work as intended on Windows. I'll see if I got some spare time this afternoon to test again on Windows but I don't think that it's blocking. Nice call on checking that out :)

bacongobbler:
LGTM from the Windows side

radu-matei:
Nice, thanks for testing, @bacongobbler!

pkg/builder/builder.go
```diff
@@ -264,7 +264,9 @@ func loadValues(ctx *Context) error {
 }
 
 func archiveSrc(ctx *Context) error {
-	contextDir, relDockerfile, err := build.GetContextFromLocalDir(ctx.AppDir, ctx.Env.Dockerfile)
+
+	dockerfilePath := filepath.Join(ctx.AppDir, ctx.Env.Dockerfile)
+	contextDir, relDockerfile, err := build.GetContextFromLocalDir(ctx.AppDir, dockerfilePath)
```
radu-matei: Before this change, the second parameter passed to `build.GetContextFromLocalDir` could have been the empty string, and the default Dockerfile would have been `Dockerfile`.

Now, we surely pass at least the application directory, so it will never be the empty string - so Docker expects that path to be an actual Dockerfile, resulting in #808

PR #800, opened 2018-05-30
Use non-onbuild base images in packs

### LINKED ISSUE #745: Packs use onbuild Docker images, which are Deprecated
Per https://github.com/docker-library/official-images/issues/2076 the `onbuild` variants of docker images are deprecated and should not be used. Several packs use this variant and should be rewritten to use the non-onbuild variant instead.

bjornmagnusson:
Closes #745


bacongobbler:
thanks @bjornmagnusson

PR #809, opened 2018-06-04
Bumped to dotnet Docker images to 2.1

dave-hillier:


dave-hillier:
These are the correct docker images, but it doesnt seem to be working. Maybe these images are broken in some way.

dave-hillier:
I found the solution in this thread: https://github.com/dotnet/coreclr/issues/13542 but I'm not sure if this is a change of behaviour or something for my app. I will try with a basic example later.

bacongobbler:
thanks @dave-hillier!

PR #831, opened 2018-07-05
[DEP] Adding proposal for multiple docker builds

danmikita:


ehotinger:
IMO both sync and parallel modes should be supported from a tool like Draft. Not all builders support parallelism because of locks around state files.

radu-matei:
Hey, @danmikita!
First of all, thanks for your proposal!

So if I understand this correctly, the distinction between the two modes would be done by where the `chart` field is - in the `global` Draft file or in a `module`?

Is there an additional way of handling this that I missed from your PR, and would it make sense to have an additional way of making the distinction between the two modes?

danmikita:
Hey @radu-matei,
We had a quick conversation at the end of your talk @ Kubecon regarding this feature.  

I re-read my old DEP from this PR and the problem it is trying to solve is still very real. I approached this initially from a maven perspective and how it uses pom files, inheritance, and execution order. With that being said, adding that kind of implementation can be confusing/overwhelming and overly complex. Perhaps there are simpler approaches to supporting this functionality with a single Draft.toml file. 

I am toying with the idea of just a list in the toml file that  1.) points to the sub-module and 2.) defines the execution order (if needed). There is still the question of modifying the chart with multiple image build ids, as well as supporting multiple charts.

I am definitely open to other ideas about how this gets implemented and would love to dialog about it. I'd like to have a rough consensus/direction before I start implementation.

PR #837, opened 2018-07-10
Updated maven pack to java 10

### LINKED ISSUE #726: Update tags for Maven and OpenJDK image in default Dockerfile
Default Dockerfile generated for a Java sample application is:

```
FROM maven:3.5-jdk-8 as BUILD

COPY . /usr/src/app
RUN mvn --batch-mode -f /usr/src/app/pom.xml clean package

FROM openjdk:8-jdk
ENV PORT 4567
EXPOSE 4567
COPY --from=BUILD /usr/src/app/target /opt/target
WORKDIR /opt/target

CMD ["/bin/bash", "-c", "find -type f -name '*.jar' | xargs java -jar"]
```

Some suggested improvements:

. The suggested tag for `maven` should be `3-jdk-10`
. The second stage of the build should use JRE instead of JDK. So, the suggested tag would be `jre-slim` instead `8-jdk`. This will create a leaner image.

Identified during hacking with @michelleN at #KubeCon

bjornmagnusson:
Fixes #726

### DIFF
--- a/packs/java/Dockerfile
+++ b/packs/java/Dockerfile
@@ -1,9 +1,9 @@
-FROM maven:3.5-jdk-8 as BUILD
+FROM maven:3-jdk-10 as BUILD
 
 COPY . /usr/src/app
 RUN mvn --batch-mode -f /usr/src/app/pom.xml clean package
 
-FROM openjdk:8-jdk
+FROM openjdk:10-jre-slim
 ENV PORT 4567
 EXPOSE 4567
 COPY --from=BUILD /usr/src/app/target /opt/target


maver1ck:
@bjornmagnusson 
Should we have one pack for java ? Java 8 is still very popular. So maybe java "classic" and new ?

birdayz:
i think packs have to much smarter:
- Detect programming language version
- Detect build tool
- Maybe support for specific frameworks

otherwise this debate might never end :) there's just not 'one' way to do something for each programming language.

bjornmagnusson:
@maver1ck The recommendations of upgrading maven/java comes from Java Champion / Docker Captain @arun-gupta. 

I think adding additional packs for the combinations of gradle/maven/java would be quite hard to maintain in the long run, there would be quite some packs, and if specific versions is required it is very much possible to create your own packs.

Personally, I think the included packs should serve as a reasonable starting point. Currently I do believe that Java 10 is that starting point for the moment, being the only Java version that is fully container aware.

For the upgrade to Java10, this is quite necessary when it comes to containers as Java below 9/10 are not container aware making it quite difficult to set reasonable constraints (like memory and cpu), without adding extra logics and parameters into the java application. Just saying, using java 8 you still will need to adapt the Dockerfile to add extra stuff to make it work in a reasonable manner in K8S.  

However, I agree with @birdayz, that the packs intelligence would need to become smarter in order to handle all these permutations of versions and so on.

Issue #835, opened 2018-07-09
support env vars in all tasks [ currently only supported in post-deploy tasks ]


michelleN:
Currently, you can pass environment variables in tasks via the `$SOME_ENV_VAR` pattern and it will get evaluated in a post-deploy task. This should be supported in all types of tasks.

PR #840, opened 2018-07-15
Allow $VARIABLES to exist in any task argument

alindeman:
I took a look at #835 because it seemed relatively simple and a fun way to read some code 🙂

As far as I can tell, `$VARIABLES` _are_ evaluated in all kinds of tasks because both pre-up and post-delete tasks [call through `executeTask`, which in turn calls `evaluateArgs`, which does the `$VARIABLES` expansion](https://github.com/Azure/draft/blob/00cf34da7b70ac9733809115a9be5c7dc5d75c5f/pkg/tasks/tasks.go#L81).

However, there are a few edge cases which I attempt to address in this PR:
* `$VARIABLES` are not expanded unless `$` is the first character in the space-separated argument: `task = "echo /home/$USER"` would not be expanded because a space-separated arg doesn't begin with `$`
* `$VARIABLES` are assumed to take up the entire space-separated argument: `task = "curl -u $USER:$PASS"` would not be expanded properly because the code would end up running `os.Getenv("USER:$PASS")` instead of expanding `$USER` and `$PASS` individually.

This proposed fix is relatively conservative. It handles expanding `$VARIABLES` in any part of the argument and allows them to be escaped as `$$VARIABLES` and `\$VARIABLES`. It does not attempt to implement other shell expansions, like `${VARIABLES}` or `${VARIABLES:-default}`.

It also keeps the current version of arg splitting (`strings.Split(task, " ")`) even though this kind of split doesn't correctly handle all cases that I think users might expect (e.g., `echo 'foo bar baz'` would be split as `[]string{"echo", "'foo", "bar", "baz'"}` instead of `[]echo{"echo", "foo bar baz"}`). There appear to be some high-quality shell splitting libraries out there (e.g., [buildkite/shellwords](https://github.com/buildkite/shellwords)) but I decided not to add these without some discussion first.

Finally, I updated dep-008.md to replace `post-delete` with `cleanup`. The original proposal specified that post-delete tasks should appear in the `post-delete` section, but the implementation only handles `cleanup`. I decided to update the proposal with the implementation.

Looking forward to hearing some feedback. Thanks for the fun Sunday morning project!

### DIFF
--- a/cmd/draft/delete.go
+++ b/cmd/draft/delete.go
@@ -108,7 +108,7 @@ func Delete(app string) error {
 			return err
 		}
 	} else {
-		if _, err = taskList.Run(tasks.PostDelete, ""); err != nil {
+		if _, err = taskList.Run(tasks.DefaultRunner, tasks.PostDelete, ""); err != nil {
 			return err
 		}
 	}
--- a/cmd/draft/up.go
+++ b/cmd/draft/up.go
@@ -155,7 +155,7 @@ func (u *upCmd) run(environment string) (err error) {
 			return err
 		}
 	} else {
-		if _, err = taskList.Run(tasks.PreUp, ""); err != nil {
+		if _, err = taskList.Run(tasks.DefaultRunner, tasks.PreUp, ""); err != nil {
 			return err
 		}
 	}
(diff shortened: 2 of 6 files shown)


michelleN:
@alindeman - we may go back to using `post-delete` instead of `cleanup` just because it's clearer what is happening. Would you mind opening an issue about using a shell splitting library? 

Thank you again. 🎉

PR #851, opened 2018-08-02
Fix create permissions

sam-doordash:
This includes a major logic change and the appropriate tests. The big change: the Pack struct's File data is now of type map[string]PackFile instead of map[string]io.ReadCloser. This allows Pack to have both the data from a file, as well as its file permissions. Everything using Pack has been updated to use the new structure.

### DIFF
--- a/pkg/draft/pack/pack_test.go
+++ b/pkg/draft/pack/pack_test.go
@@ -24,7 +24,7 @@ cleanup-task = "echo cleanup"
 
 func TestSaveDir(t *testing.T) {
 	dockerPerm := os.FileMode(0666)
-	tasksPerm := os.FileMode(0777)
+	tasksPerm := os.FileMode(0655)
 	p := &Pack{
 		Chart: &chart.Chart{
 			Metadata: &chart.Metadata{
--- a/pkg/draft/pack/pack_test.go
+++ b/pkg/draft/pack/pack_test.go
@@ -2,7 +2,6 @@ package pack
 
 import (
 	"bytes"
-	"fmt"
 	"io/ioutil"
 	"k8s.io/helm/pkg/proto/hapi/chart"
 	"os"
@@ -55,8 +54,7 @@ func TestSaveDir(t *testing.T) {
 		}
 	}
 	if fInfo.Mode() != dockerPerm {
-		fmt.Println("DockerFile perms different")
-		t.Fail()
+		t.Errorf("DockerFile perms different. Expected %s, but got %s", dockerPerm, fInfo.Mode())
 	}
 
 	tasksPath := filepath.Join(dir, TargetTasksFileName)
@@ -69,8 +67,7 @@ func TestSaveDir(t *testing.T) {
 		}
 	}
 	if fInfo.Mode() != tasksPerm {
-		fmt.Println("Tasks file perms different")
-		t.Fail()
+		t.Errorf("Tasks file perms different. Expected %s, but got %s", tasksPerm, fInfo.Mode())
 	}
 
 	data, err := ioutil.ReadFile(tasksPath)
--- a/pkg/draft/pack/pack_test.go
+++ b/pkg/draft/pack/pack_test.go
@@ -2,13 +2,13 @@ package pack
 
 import (
 	"bytes"
+	"fmt"
 	"io/ioutil"
+	"k8s.io/helm/pkg/proto/hapi/chart"
 	"os"
 	"path/filepath"
 	"testing"
-	"k8s.io/helm/pkg/proto/hapi/chart"
-	"fmt"
-	)
+)
 
 const testDockerfile = `FROM nginx:latest
 `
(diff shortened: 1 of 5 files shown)


michelleN:
hey @sam-doordash - thanks for your contribution. Looks like the the linters are not passing. Would you mind seeing if you can fix those formatting issues? In the meantime, I'll review the code. Cheers.

sam-doordash:
@michelleN - gofmtd the necessary files

sam-doordash:
@michelleN - got all the tests working for windows now, too. The behavior is slightly different for windows v linux/mac (it only sets a single read/write permission for all users/groups based on the linux owner permissions and doesn't set execute in windows)

sam-doordash:
@michelleN - any updates on merging this + https://github.com/Azure/draft/pull/848?

michelleN:
Thanks for all the updates @sam-doordash. I'll put this on my list of items to review today and get back to you this afternoon.

michelleN:
Ah wasn't able to get to this last night but it's a priority for today!

pkg/draft/pack/pack_test.go
```diff
@@ -2,13 +2,12 @@ package pack
 
 import (
 	"bytes"
-	"io"
 	"io/ioutil"
+	"k8s.io/helm/pkg/proto/hapi/chart"
```
radu-matei: Could you please also `goimports`, so all imports are ordered? Thanks!

pkg/draft/pack/pack_test.go
```diff
@@ -24,15 +23,19 @@ cleanup-task = "echo cleanup"
 `
 
 func TestSaveDir(t *testing.T) {
+	dockerPerm := os.FileMode(0664)
+	winDockerPerm := os.FileMode(0666)
+	tasksPerm := os.FileMode(0644)
+	winTasksPerm := os.FileMode(0666)
```
radu-matei: Surely nothing that would prevent merging this PR, but I think would help if we had the permissions as constants, with expressive naming - this is something we should use across Draft (as there are also other uses of permission literals).

But not necessarily concerning this PR.

pkg/draft/pack/pack_test.go
```diff
@@ -45,24 +48,36 @@ func TestSaveDir(t *testing.T) {
...
-	_, err = os.Stat(filepath.Join(dir, dockerfileName))
+	fInfo, err := os.Stat(filepath.Join(dir, dockerfileName))
 	if err != nil {
 		if os.IsNotExist(err) {
 			t.Errorf("Expected %s to be created but wasn't", dockerfileName)
 		} else {
 			t.Fatal(err)
 		}
 	}
+	if fInfo.Mode() != dockerPerm && runtime.GOOS != "windows" {
```
radu-matei: As there are multiple places in this test where we test for the runtime, would it make sense to use conditional compilation (`// +build windows`)?

radu-matei: But as with the previous comment, not necessarily blocking this PR.

pkg/draft/pack/pack_test.go
```diff
@@ -45,24 +48,36 @@ func TestSaveDir(t *testing.T) {
...
 		} else {
 			t.Fatal(err)
 		}
 	}
+	if fInfo.Mode() != tasksPerm && runtime.GOOS != "windows" {
+		t.Errorf("Tasks file perms different. Expected %s, but got %s", tasksPerm, fInfo.Mode())
+	}
+	if fInfo.Mode() != winTasksPerm && runtime.GOOS == "windows" {
+		t.Errorf("Tasks file perms different. Expected %s, but got %s", tasksPerm, fInfo.Mode())
+	}
```
michelleN: 👍

pkg/draft/pack/pack_test.go
michelleN: This is in a test and not necessarily representative of what the file permissions would be for these files. Do you still think a constant would be helpful here or do you think we can leave these as variable? @radu-matei

pkg/draft/pack/pack_test.go
sam-doordash: I've never used conditional compilation before, but that sounds like it would be a really clean solution

sam-doordash:
@michelleN - for the test `TestSaveDirDockerfileExistsInAppDir`, the only change in logic I made was the object type in the pack to reflect the changes in the rest of the code. I think it is verifying that the Dockerfile can be overwritten, but I'm not 100% myself.

pkg/draft/pack/pack_test.go
radu-matei: Hey!
You can find an example with conditional building in tests in this repo - https://github.com/Azure/draft/tree/master/pkg/draft/draftpath

Take a look at that, and if you have any questions I'd be glad to help 😄

michelleN:
yea @sam-doordash - I agree on that test is ensuring that the file can be overwritten which shouldn't be the behavior and needs a closer look. I opened up an issue to remind myself to go do that #866.

PR #857, opened 2018-08-18
Packs: corrected gradle docker image

bjornmagnusson:
Corrects docker image in the gradle pack.
Fixes in file permissions and the way the app is ran


bjornmagnusson:
@michelleN This should probably be in the 0.16 release as well

PR #858, opened 2018-08-18
Gradle example app, migrated from example-java

### LINKED ISSUE #849: Each pack should have an example app (gradle missing)
There should be an example app available for every `pack`. This to enable easy verification of changes to packs and to be able to show how an pack should be used.

Realized with #838 that there is no example app available for gradle.

@michelleN I think this is the reason for your "error" when trying the gradle pack with example-java (only maven build script) :) .

bjornmagnusson:
Fixes #849 
Stole the app in example-java written for maven, migrated it to Gradle


michelleN:
I was probably getting that error because I wasn't using your updated Dockerfile!

PR #862, opened 2018-09-08
Remove onbuild from docs

### LINKED ISSUE #861: Getting started docs indicate use of onbuild image
docs/getting-started.md still states `This Dockerfile harnesses the python:onbuild image` which is no longer true

bjornmagnusson:
As docker images used in bundled packs does not use onbuild images anymore, the docs should reflect that

Fixes #861

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -23,7 +23,7 @@ $ ls -a
 .draft-tasks.toml Dockerfile        charts/           requirements.txt
 ```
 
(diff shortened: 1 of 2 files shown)


radu-matei:
Thanks for your contribution, @bjornmagnusson!

I think we should also remove the `onbuild` references form the test data, but this is more of a note for us that shouldn't stop this PR from being merged.

Thanks!

PR #647, opened 2018-04-06
Dot 13: updated documentation 

squillace:
for rc1 .13

PR #870, opened 2018-10-02
Fix dead link to installation guide and update helm repository link

chrischdi:
Fixes a dead link to the install-minikube.md guide which was previously named `install.md` and got renamed in #647.

### DIFF
--- a/docs/contributing/hacking.md
+++ b/docs/contributing/hacking.md
@@ -85,10 +85,10 @@ $ make clean
 
 
 [docker]: https://www.docker.com/
-[install]: ../install.md
+[install]: ../install-minikube.md
 [git]: https://git-scm.com/
 [go]: https://golang.org/
-[helm]: https://github.com/kubernetes/helm
+[helm]: https://github.com/helm/helm
 [Homebrew]: https://brew.sh/
 [Kubernetes]: https://github.com/kubernetes/kubernetes
 [minikube]: https://github.com/kubernetes/minikube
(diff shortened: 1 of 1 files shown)


docs/contributing/hacking.md
```diff
@@ -85,10 +85,10 @@ $ make clean
 
 
 [docker]: https://www.docker.com/
-[install]: ../install.md
+[install]: ../install-minikube.md
```
radu-matei: Would this - https://github.com/Azure/draft/blob/master/docs/quickstart.md#install-and-configure-helm be a better section to point for installing Helm?

docs/contributing/hacking.md
```diff
@@ -85,10 +85,10 @@ $ make clean
...
 [git]: https://git-scm.com/
 [go]: https://golang.org/
-[helm]: https://github.com/kubernetes/helm
+[helm]: https://github.com/helm/helm
```
radu-matei: 👍

docs/contributing/hacking.md
chrischdi: Thanks for your review.

I think you're right. 

I think the text where the link is referenced should get edited:
At line 13 there is:
`- [helm][], using the same version as recommended in the [installation guide][install].`

There is no recommended version on neither one of the two references if I'm right?

I propose to change this to:
`- [helm][], see the [quickstart guide][quickstart] for installing helm`
Or to add a general reference at this section to the quickstart guide?

What do you think?

PR #871, opened 2018-10-02
feat(add): disable emoji flag

### LINKED ISSUE #304: flag to enable or disable emojis
See #294 for context

sendilkumarn:
fix #304 

r? @radu-matei

### DIFF
--- a/pkg/cmdline/cmdline.go
+++ b/pkg/cmdline/cmdline.go
@@ -171,7 +171,7 @@ func concatStrAndEmoji(text string, emoji string, disableEmoji bool) string {
 	if !disableEmoji {
 		concatStr.WriteString(emoji)
 	}
-	concatStr.String()
+	return concatStr.String()
 }
 
 func consoleSupportsColor() bool {
--- a/pkg/cmdline/cmdline.go
+++ b/pkg/cmdline/cmdline.go
@@ -1,9 +1,9 @@
 package cmdline
 
 import (
-	"bytes"
 	"fmt"
 	"runtime"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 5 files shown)


pkg/cmdline/cmdline.go
```diff
@@ -156,14 +157,21 @@ func progress(cli *cmdline, app, desc string, codes <-chan builder.SummaryStatus
...
+	return fmt.Sprintf("%s: %s", red(msg), concatStrAndEmoji("FAIL", " ❌ ", disableEmoji))
+}
+
+func concatStrAndEmoji(text string, emoji string, disableEmoji bool) string {
+	var concatStr bytes.Buffer
+	concatStr.WriteString(text)
+	if !disableEmoji {
+		concatStr.WriteString(emoji)
+	}
+	concatStr.String()
```
radu-matei: This should probably return.

pkg/cmdline/cmdline.go
```diff
@@ -156,14 +157,21 @@ func progress(cli *cmdline, app, desc string, codes <-chan builder.SummaryStatus
...
 
-func failStr(msg string) string {
-	const fail = "FAIL " + "❌"
-	return fmt.Sprintf("%s: %s", red(msg), fail)
+func failStr(msg string, disableEmoji bool) string {
+	return fmt.Sprintf("%s: %s", red(msg), concatStrAndEmoji("FAIL", " ❌ ", disableEmoji))
+}
+
+func concatStrAndEmoji(text string, emoji string, disableEmoji bool) string {
+	var concatStr bytes.Buffer
```
bacongobbler: more of an FYI, but you can use a [`strings.Builder`](https://golang.org/pkg/strings/#Builder) here instead for more efficient memory usage.

pkg/cmdline/cmdline.go
```diff
@@ -156,14 +157,21 @@ func progress(cli *cmdline, app, desc string, codes <-chan builder.SummaryStatus
...
+	concatStr.WriteString(text)
+	if !disableEmoji {
```
itowlson: I find the use of `disableEmoji` throughout a bit hard to follow, and this double negative is I think why - there is an extra cognitive load of 'and pass whether we're _not_ showing emoji' and 'if we're **not** not showing emoji...".  Consider a positive name `displayEmoji` and a positive condition check.

pkg/cmdline/cmdline.go
sendilkumarn: Thanks, it was introduced in 1.10 right? Gotcha

pkg/cmdline/cmdline.go
sendilkumarn: Agree to it 👍  changed it

radu-matei:
Now that #874 is merged, a rebase from `master` should make the CI go green 😄

sendilkumarn:
@radu-matei Done 👍

radu-matei:
Manually tested again, LGTM.

radu-matei:
Thanks for your contribution, @sendilkumarn!
Let us know when you want to pick up something else here, we'd be more than happy to help 😄

PR #892, opened 2018-10-27
Remove onbuild from testdata

### LINKED ISSUE #890: Remove onbuild from testdata
As the Docker images being used for the packs does not use the onbuild versions, the onbuild should be removed from the testdata in various places as well.
This is sort of an left-over from #862

kshitij10496:
This PR aims at removing the `onbuild` tags from the Dockerfiles in the testdata.
Resolves #890 

cc @bjornmagnusson


bjornmagnusson:
@kshitij10496 Would you mind taking a look at the test failures?

kshitij10496:
Thank you for your patience. I'm looking into it! 😅

Issue #366, opened 2017-09-27
Support for Windows Containers


radu-matei:
With the new announcement for Windows Server 1709 you will be able to mount named pipes in containers, which will allow you to build Docker images inside a container by mounting `\\.\pipe\docker_engine`.

(See announcements [in the Docker blog post](https://blog.docker.com/2017/09/docker-windows-server-1709/) and on [Windws Insider 16278 blog post](https://blogs.windows.com/windowsexperience/2017/09/05/announcing-windows-server-insider-preview-build-16278))

If I am not missing anything, this should allow the `draft` server to run in a Windows container and create the image, right?

I would love to take a look at this - the problem at the moment is that I haven't worked with Windows agents in Kubernetes, and apart from creating a hybrid cluster (Linux + Windows nodes) with `az` and `acs-engine` I don't know how to configure the Windows part manually - manually because you need the latest Windows Server preview to have named pipe support.

The approach that came in mind is the following: Azure VM with the default Server 2016 on a v3 instance that supports nested virtualization, create there a Linux + a Windows Server Insider build 16278 and configure the cluster there.

(Putting this here because maybe someone will get to try this before me and hopefully document the process :D or can point me to a resource for Windows agents in Kubernetes)

Is there any other blocker for Draft on Windows Server that I missed?

Thanks!

bacongobbler:
If you can get a working kubernetes cluster on Windows Server and can point us to a blog, maybe one of us can give it a whirl. I don't see the issue if the endpoint is Yet Another Kubernetes Endpoint.

JasonvanBrackel:
@radu-matei Have you spent any time with this.  If so I'd love to hear your progress.  If not, I may take a stab at this.  I'm a member of sig-windows and familiar with Windows Server Container in K8s.

jsturtevant:
I looked into this a little. The docs state that the local files are shipped up to `draftd`, the server side component and the image is built there. Looking briefly [at the code](https://github.com/Azure/draft/blob/master/pkg/draft/server.go#L213), it does look like the server side component (`draftd`) does the building of the image.  

It might it be possible to get this to run on a windows server and have the windows specific builds be done there?  Then there would need to be some updates to make sure the correct files are built on the right node type.  Having two server-side components might not be ideal.

Would any one from the Draft team be able to give some guidance on how this might be designed to work for windows (or mixed cluster)?  Am I way off in how this works?

bacongobbler:
[...]

PR #893, opened 2018-10-28
feat(pack): Add dotnet framework WCF example for Windows

sylus:
Hi there ^_^

This was discussed a bit in #366 and wasn't sure the correct way to go about this so added a `win` flag and a .NET Framework WCF example for windows.


PatrickLang:
I was just about to start one of these!

Issue #765, opened 2018-05-18
re-do README


michelleN:
Based on various feedback from newcomers to Draft, we've decided we need to take another look at the README.md file. This is a meta-ticket to track re-working our main draft README

Items include:
- [ ] #610 
- [ ] gifs instead of the asciiema

Fresh perspectives welcome.

PR #894, opened 2018-10-28
Updated README to include gif instead of asciinema

jQubz:
Resolved part of #765

Couldn't find a better way to host such a large gif file and cannot embed an iframe form gfycat. This is the best way I could think of.


cdrage:
@jQubz https://github.com/asciinema/asciicast2gif works pretty well. I took a 30-second asciinema and it converted to a 800kb file :+1:

Issue #883, opened 2018-10-22
Use conditional compilation for tests


radu-matei:
There are [a bunch of places](https://github.com/Azure/draft/search?l=Go&q=runtime.goos) where we have if statements to check for the runtime OS. 

A really clean solution would be to use conditional compilation - and if anyone wants to take a stab at this, here's a [really useful resource for conditional compilation](https://dave.cheney.net/2013/10/12/how-to-use-conditional-compilation-with-the-go-build-tool).

bege13mot:
Hi,

Hope my PR will be helpful.

PR #895, opened 2018-10-28
Use conditional compilation for tests

bege13mot:
https://github.com/Azure/draft/issues/883
Signed-off-by: bege13mot <bege13mot@gmail.com>

### DIFF
--- a/pkg/draft/pack/pack_windows_test.go
+++ b/pkg/draft/pack/pack_windows_test.go
@@ -1,5 +1,3 @@
-// +build !windows
-
 package pack
 
 import (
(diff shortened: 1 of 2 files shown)


pkg/draft/pack/pack_windows_test.go
```diff
@@ -0,0 +1,120 @@
+package pack
```
bacongobbler: do we need a conditional build flag for windows, or is Go smart enough about the `_windows` filename that it implies the windows build tag? I recall reading about that somewhere...

bege13mot: Hi @bacongobbler ,
As I can understand if we try to build it on a windows machine, there is no need to specify window build tag.

> (example: source_windows_amd64.go) where GOOS and GOARCH represent any known operating system and architecture values respectively, then the file is considered to have an implicit build constraint requiring those terms (in addition to any explicit constraints in the file).

https://golang.org/pkg/go/build/

Issue #547, opened 2018-02-26
Draft up should allow --dry-run


thedrow:
Debugging templates is currently not as easy as it should be.
Helm provides a `--dry-run` flag for its create action that allows users to understand what will be created.
Currently draft lacks such mechanism for `draft up`.

michelleN:
Oh very interesting! Could you expand a little more on what you'd like to see? When I read this, I think an end user might want something like this:
```console
$ draft up --dry-run
Draft will bundle up this directory and send it to the draft server in this cluster: <insert cluster name> where it will 1. Create a Docker image: <insert image name:tag> and send it to this container registry: < insert registry name> 2. Deploy <insert image name:tag> using the helm chart provided in `charts/`
```
or 
$ draft up --dry-run --verbose
Draft will bundle up this directory and send it to the draft server in this cluster: <insert cluster name> where it will 1. Create a Docker image: <insert image name:tag> and send it to this container registry: < insert registry name> 2. Deploy <insert image name:tag> using the helm chart provided in `charts/`. See what that chart deployment would look like below:
<insert rendered chart from helm install --dry-run --debug>
```

Is that what you were thinking as well or were you wanting something different?

thedrow:
The second one is what I was expecting.

janowsiany:
Hey @michelleN is this still valid? I would like to work on this.

radu-matei:
Hey, @janowsiany!
Sure, this is something we want - let us know if we can help in any way.

Thanks!

PR #898, opened 2018-10-30
Introduce dry-run flag for up command

janowsiany:
Addresses idea raised in #547 
`--verbose` still needs to be added to extend the output with helm charts


janowsiany:
ping @michelleN @radu-matei

bacongobbler:
there seems to be a lot of duplicate code shared here between `pkg/builder` and this PR. Is there a way we could refactor this into a more clean interface?

radu-matei:
I really like @bacongobbler's idea - if we could reuse all that logic it would be great!

Issue #856, opened 2018-08-11
Feature Request: Add ability to specify array of helm chart values files


Dm3Ch:
It would be great to have ability to specify array of values files for helm chart in draft toml config.

For example it could be needed for working on application in public git repository and keep `values.yaml` with not sensitive params and have second values files with sensitive params ignored by git. 
So people could see in repo properties required by chart but not see sensitive data

maximbaz:
Related: #818

radu-matei:
Hey, @zoidbergwill!
This looks great! Let us know if you have any questions, we'd be happy to help :)

veeramarni:
Wondering if there is any progress on this issue. It is quite a reasonable requirement when we have pipelines to deploy already built docker images to another environment without having to build the images again.

zoidbergwill:
Whoops, I disappeared. The last month has been a bit busier then expected. Will work on this today.

maximbaz:
FYI, @zoidbergwill has already sent out an initial PR (thanks!): #899

zoidbergwill:
Please feel free to comment on it. I'll work on adding test coverage this weekend.

Issue #818, opened 2018-06-08
customizable ValuesfileName


omnilinguist:
Noticed that in the file https://github.com/Azure/draft/blob/4a298770ff4484e5737fabf4bb9abc54b6b72e7c/pkg/draft/pack/pack.go#L19-L20, there is a section
```
// ValuesfileName is the default values file name.
ValuesfileName = "values.yaml"
```
However, this constant doesn't seem to be used anywhere else in the codebase, which makes me wonder how it gets used by draft.
The answer seems to be in https://github.com/kubernetes/helm/blob/cf3ded91f2143fed6850dee70630c1049bf937ec/pkg/chartutil/create.go#L31-L32, but whereas there is a way to configure this in helm via `helm ... -f values_env.yaml`, the draft wrapper doesn't seem to expose this functionality.

A more general version of this, of course, would be to have a more direct interface with the underlying helm commands that get invoked for advanced users.

Use case for this (the shrewd reader may again sense a CI/CD flavour here) is that if an application's chart has multiple values YAMLs for different environments (staging, production), it would be possible to toggle easily between the two rather than having to have complex toggling constructs in the Helm template files.

bacongobbler:
Probably just dead code. We can clean that up.

maximbaz-microsoft:
We have a very similar setup: `values.yaml` in root of a chart with common properties, and a few `values.yaml` files in `environments/<env>/` folders, and some secrets that are always being passed as a param to `helm`.

It seems `draft` can read the parent `values.yaml`, and it has an interface to configure secrets (in draft.toml in `set` list), but there is no way to tell draft to _also_ look in `environments/dev/values.yaml` just like I would do it with `helm ... -f environments/dev/values.yaml`.

maximbaz:
We have a very similar setup: `values.yaml` in root of a chart with common properties, and a few `values.yaml` files in `environments/<env>/` folders, and some secrets that are always being passed as a param to `helm`.

It seems `draft` can read the parent `values.yaml`, and it has an interface to configure secrets (in draft.toml in `set` list), but there is no way to tell draft to _also_ look in `environments/dev/values.yaml` just like I would do it with `helm ... -f environments/dev/values.yaml`.

zoidbergwill:
@bacongobbler from what I can see these constants aren't used anywhere in draft, should we keep them?

```
-       // ChartfileName is the default Chart file name.
-       ChartfileName = "Chart.yaml"
-       // ValuesfileName is the default values file name.
[...]

PR #899, opened 2018-10-31
Support specifying custom values files

zoidbergwill:
Helm has code for this here: https://github.com/helm/helm/blob/master/cmd/helm/install.go#L358

It would probably be a better approach to try move their code for it to
k8s.io/helm/pkg/chartutil, and then use that here. For now I just copied out
the necessary bits, because their version supports things like custom protocols,
like values files from ftp or http, which I don't think we need. Though
matching what helm supports is probably a better idea.

TODO: I still need to add tests for this.

Closes https://github.com/Azure/draft/issues/818
Closes https://github.com/Azure/draft/issues/856

Signed-off-by: William Stewart <zoidbergwill@gmail.com>

### DIFF
--- a/pkg/builder/builder.go
+++ b/pkg/builder/builder.go
@@ -24,6 +24,7 @@ import (
 	"github.com/docker/docker/pkg/fileutils"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/net/context"
+	yaml "gopkg.in/yaml.v2"
 	"k8s.io/api/core/v1"
 	apiErrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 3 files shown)


pkg/builder/builder.go
```diff
@@ -250,14 +251,67 @@ func loadArchive(ctx *Context) (err error) {
 	return nil
 }
 
+// Merges source and destination map, preferring values from the source map
+func mergeValues(dest map[string]interface{}, src map[string]interface{}) map[string]interface{} {
```
radu-matei: cc @bacongobbler, @michelleN 
Are there any other projects you are aware of that re-implement the merging of value files from Helm?

Would it make sense to expose this in the Helm `pkg`?

maximbaz: Google's `skaffold`, for example, just [calls `helm` binary directly](https://github.com/GoogleContainerTools/skaffold/blob/3c275201e8e1f7cf4522897394467aa8f6bc34c6/pkg/skaffold/deploy/helm.go#L109-L112) and [passes all `valuesFiles` as arguments](https://github.com/GoogleContainerTools/skaffold/blob/3c275201e8e1f7cf4522897394467aa8f6bc34c6/pkg/skaffold/deploy/helm.go#L213-L215) to it, i.e. it doesn't deal with merging them at all — but even if this is the first occurrence, I think it makes sense to expose this in the Helm.

zoidbergwill: Should I make a PR for helm? Would `chartutil` be the right package for it?

bacongobbler: I'm pretty sure this already exists in `pkg/chartutil` called `MergeInto`: https://github.com/helm/helm/blob/403af2c3895e4fa0e04713d94135bf3438d9ce9c/pkg/chartutil/values.go#L97-L111

Probably would be good to use that instead

bacongobbler: Usage: https://github.com/helm/helm/blob/82d01cb3124906e97caceb967a09f2941d6a392d/pkg/chartutil/values_test.go#L521-L540

zoidbergwill: From what I could see, parts where supported by `chartutil` already, but not other parts, will take a look at refining it, whether it's in Helm or here.

cws-credsimple:
Any word when, or if, this will be merged? This is a must-have feature for my particular use case.

ofekih:
Is there any way I can help get this pull request back alive? I can contribute to the code if necessary, but like @cws-credsimple, this is very important for my use case

PR #906, opened 2018-12-10
fix ctrl c not registering issue

michelleN:


faraazkhan:
Just tested this, works like a charm! 
https://asciinema.org/a/VfqIizWrMXVXR4neelM9Lxk5V

Thanks for the quick fix!

PR #928, opened 2019-03-02
Enable color output on windows

corneliusweig:
Color output on PowerShell needs to be explicitly enabled. Usually logrus does that, but only when printing a log message. To be on the safe side, do this explicitly during the CLI initialization. The implementation is based on how logrus does it.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -615,6 +615,14 @@
   version = "1.0.4"
 
 [[projects]]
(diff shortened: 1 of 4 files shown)


corneliusweig:
@itowlson Please let me know if you have recommendations how to improve this.

itowlson:
LGTM, thanks @corneliusweig!

Issue #934, opened 2019-05-09
Go Modules support


nicmr:
Hi, I'm currently working on updating draft to work with [go modules](https://blog.golang.org/using-go-modules).
This is the current progress:
- [x] Building and running draft outside the gopath
- [ ] Using draft code in /pkg/ as a go modules dependency

# 1. Basic migration
Following the [official documentation](https://github.com/golang/go/wiki/Modules#how-to-define-a-module), I went through the following process:
```shell
# clone outside GOPATH
$ git clone https://github.com/Azure/draft
$ cd draft
$ go mod init github.com/Azure/draft
```
go mod will then copy the dependency requirements from the existing Gopkg.lock.

```shell
$ go mod tidy
```
to clean up the dependencies , add missing ones and mark indirect dependencies.
The results are [this go.mod](https://github.com/nicmr/draft/blob/master/go.mod) and [this go.sum](https://github.com/nicmr/draft/blob/master/go.sum).

Now I can build and run draft outside the gopath.

```shell
$ go build ./...
$ go run ./cmd/draft
```
# 2. Remaining issues
There are still some issues when importing the packages in /pkg/ (or /cmd/) in a different module.

### sirupsen/logrus 
I've encountered the following issues:
```shell
$ go mod tidy
go: github.com/Sirupsen/logrus@v1.4.1: parsing go.mod: unexpected module path "github.com/sirupsen/logrus"
```
This appears to be caused by a breaking change in capitalisation of that github user (`Sirupsen` to `sirupsen`), which is a breaking change, but not changing the major semver of the package.

Usually, we should be able to handle this with the replace syntax in go.mod

```
// go.mod
replace github.com/Sirupsen/logrus v1.4.1 => github.com/sirupsen/logrus v1.4.1
```
but that doesn't work, as draft is already importing the package under its correct name in /cmd/draft.
When we use `go mod why ` to find out where the package is imported under the wrongly capitalized name, we get the following dependency graph:
```
$ go mod why github.com/Sirupsen/logrus
[...]
gitlab.com/altemista/altemistactl/cmd/draft
github.com/Azure/draft/pkg/builder
github.com/docker/docker/pkg/archive
github.com/Sirupsen/logrus
```
So this is appparently an indirect dependency and the mistake is in `docker/docker/pkg/archive`.
I'm going to try to find an elegant fix for this tomorrow.
For now, we can replace the dependency with an older version as a bandaid fix:
```
// go.mod (of the module importing draft)
[...]

PR #936, opened 2019-05-14
Go modules support

nicmr:
### Additions:

Adds the necessary files for Go 1.11 + modules support, `go.mod` and `go.sum`, generated from the current `Gopkg.lock`.

### Benefits:
- The program can be built outside of the gopath, by simply `running go build ./...`
-  [dep](github.com/golang/dep/cmd/dep) is no longer required

### Open questions:
Should  `make bootstrap` use `go mod` instead of `dep`?
- Go modules can't replace `make bootstrap`, e.g:
  - It can require [gometalinter](github.com/alecthomas/gometalinter) as a dependency, but can't run `gometalinter --install`
  - It cannot ensure git is installed
- the current `make bootstrap` can't be run outside the gopath (`dep ensure -v` fails). This limits the benefits gained from adding modules support.

### Related issue:
- #934 

### Approach:

The files were generated with the following approach:

```
$ git clone https://github.com/Azure/draft
$ cd draft
$ go mod init github.com/Azure/draft
$ go mod tidy
```


nicmr:
### Concerning the failed CircleCI build:
- CircleCI fails when installing `gometalinter`. According to the author, `--install` is not a valid installation instruction anymore( https://github.com/alecthomas/gometalinter/issues/552 ), so this might be the issue?
I could update the makefile to follow the suggested installation instructions of gometalinter instead.
Gometalinter has also since been deprecated, so switching to  [golangci-lint](https://github.com/golangci/golangci-lint) should be considered.

- CircleCI also ignores the added go modules support, because the tests are run in `$GOPATH/src` and `GO111MODULE=auto` is set. To also run the tests using go modules, the tests either have to be rerun with `GO111MODULE=on` or outside the `GOPATH`.

